PR #16, opened 2014-10-22
Rename 'Tophat2' tool to 'Tophat'.

natefoo:
Per @jgoecks the old Tophat (1) tool should be deprecated and Tophat2 should always be preferred. Renaming Tophat2 to simply "Tophat" to make it clearer for users who may not understand why the tool is named Tophat2.

The Tophat2 tool lacks color space functionality but we don't think it's likely anyone would be using it (and the Tophat 1 tool could be used for it).

### DIFF
--- a/tools/tophat2/tophat2_wrapper.xml
+++ b/tools/tophat2/tophat2_wrapper.xml
@@ -1,4 +1,4 @@
-<tool id="tophat2" name="Tophat2" version="0.6">
+<tool id="tophat2" name="Tophat" version="0.6">
     <!-- Wrapper compatible with Tophat version 2.0.0+ -->
     <description>Gapped-read mapper for RNA-seq data</description>
     <version_command>tophat2 --version</version_command>


martenson:
We are showing in a much clearer manner what is the Galaxy Tool version and what is the underlying requirements including the version of it. This is now live on Main.

@natefoo congrats on a nice discussion for a 1char PR :)

jgoecks:
Regardless of the tool version improvements that we've made in the tool UI, we should make this change. It will reduce user confusion significantly and bring us in line with the rest of the genomics community. The community does not distinguish between Tophat and Tophat2—they are different versions of the same tool, and nobody is using Tophat version 1.x anymore.

dannon:
+1.  All the version improvements are great, but there's no reason to call it "Tophat2" as far as I can tell.  At http://ccb.jhu.edu/software/tophat/index.shtml, it's just Tophat, with a version of 2.0.X.

martenson:
@jgoecks Sorry about closing this, I got carried away by the version being shown and did not check the start of the debate. I think we can rename this easily and will test it on the Test instance tomorrow.

martenson:
@jgoecks does the same apply to Bowtie2? Any other tools affected by this?

nsoranzo:
@martenson No! Bowtie and Bowtie2 are both still maintained in parallel.

martenson:
What is the difference in between tools on Main named `Tophat for Illumina` and `Tophat2`?

jgoecks:
Tophat for Illumina = Tophat v1.x, Tophat2 = Tophat v2.x Is there a reason this hasn't been merged?

martenson:
Closed via https://github.com/galaxyproject/tools-devteam/commit/dbd3530b318c37355692e45a0e2bfa4600c17c00 because of merge conflict :)

PR #48, opened 2014-12-22
Fixes for bwa-mem

nsoranzo:
Some fixes derived by https://bitbucket.org/crs4/orione-tools/src/HEAD/custom/bwa_mem/bwa_mem.xml .

### DIFF
--- a/tools/bwa/bwa-mem.xml
+++ b/tools/bwa/bwa-mem.xml
@@ -299,18 +299,18 @@
             <option value="do_not_set" selected="True">Do not set</option>
           </param>
           <when value="set">
-            <param name="k" type="integer" value="19" label="minimum seed length" help="-k; default=19"/>
-            <param name="w" type="integer" value="100" label="band width for banded alignment" help="-w; default=100"/>
-            <param name="d" type="integer" value="100" label="off-diagonal X-dropoff" help="-d; default=100"/>
-            <param name="r" type="float" value="1.5" label="look for internal seeds inside a seed longer than -k * THIS VALUE" help="-r; default=1.5"/>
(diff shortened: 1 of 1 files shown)


tools/bwa/bwa-mem.xml
```diff
@@ -212,21 +238,41 @@
...
-        <param name="ID" type="text" value="readgroup1" size="20" label="Specify readgroup ID" help="This value must be unique among multiple samples in your experiment">
+        <param name="ID" type="text" value="readgroup1" size="20" label="Read group identifier (ID)" help="This value must be unique among multiple samples in your experiment">
           <sanitizer invalid_char="">
             <valid initial="string.printable"/>
           </sanitizer>
         </param>
-        <param name="SM" type="text" value="blood" size="20" label="Specify readgroup sample name (SM)" help="This value should be descriptive">
+        <param name="SM" type="text" value="blood" size="20" label="Read group sample name (SM)" help="This value should be descriptive. Use pool name where a pool is being sequenced">
           <sanitizer invalid_char="">
             <valid initial="string.printable"/>
```
jmchilton: (@nsoranzo - I know you didn't add this but I just noticed this on your diff). Is this insecure? string.printable allows a lot of stuff through right and it seems to be just inserted into the command-line with sanitization.

bgruening: string.letters + string.digits are probably enough. I would prefer to leave it like this but only take the first 20 (+x?) characters.

nsoranzo: No idea, for our CRS4 bwa_mem wrapper I used ```<validator type="empty_field" />``` for ID and SM.

PR #49, opened 2014-12-23
Add conditional to choose between -n and -v alignment modes.

nsoranzo:
When specifying both -n and -v options, bowtie considers only the last one.
Moreover:

- show fields for -e, -l and --nomaqround options only for -n alignment
- min value for -e is 1 not 0
- possible values for -v are 0, 1, 2, 3

Tool version is not incremented because it was already in commit 8991438139974fb52c878401885dbdfb6a0a5bc0 and then not released on the Tool Shed.

All tests pass in Planemo, but I had to create a ```tool_data_table_conf.xml.test``` file and copy the ```equCab2chrM/bowtie/``` directory (8.1 Mb) from https://bitbucket.org/natefoo/galaxy-test-data/ . Is there interest in moving these test files in this repo? @jmchilton


nsoranzo:
May this be merged? @bgruening

bgruening:
Yes  thanks! Nice job!

Issue #53, opened 2015-01-13
BWA harmonization


jmchilton:
Make greater use of macros to further harmonize BWA and BWA-mem wrappers - including read group handling (see comments on #52).

kaktus42:
fixed current RG handling in BWA: #179

bernt-matthias:
I guess this happened at IUC

PR #54, opened 2015-01-16
Harmonize read group in BWA wrappers, see #53. Also add specific tests.

nsoranzo:
In particular this will add many read group params to ```bwa.xml```.


bgruening:
Looks good to me. I will merge it tomorrow if no one complains.

PR #56, opened 2015-01-27
Fix bugs and typos

jsh58:
Fixed numerous bugs and typos.  Added option to produce SAM output rather than sorted BAM.

### DIFF
--- a/tools/bowtie2/bowtie2_wrapper.xml
+++ b/tools/bowtie2/bowtie2_wrapper.xml
@@ -139,7 +139,6 @@
             #end if
             
             #if str( $analysis_type.other_options.other_options_selector ) == "yes":
-                ${analysis_type.other_options.reorder}
                 ${analysis_type.other_options.non_deterministic}
                 --seed "${analysis_type.other_options.seed}"
             #end if
@@ -149,11 +148,7 @@
         #end if    
         
         ## view/sort and output BAM file
-        #if ( str( $analysis_type.analysis_type_selector ) != "full" or str( $analysis_type.sam_opt ) != "true" ):
-          | samtools view -Su - | samtools sort -o - - > $output_1
-        #else
-          > $output_2
-        #end if
+        | samtools view -Su - | samtools sort -o - - > $output
 
         ## rename unaligned sequence files
         #if $library.type == "paired" and $output_unaligned_reads_l and $output_unaligned_reads_r:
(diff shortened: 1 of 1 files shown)


tools/bowtie2/bowtie2_wrapper.xml
```diff
@@ -189,10 +195,10 @@
...
+                        <param name="X" type="integer" value="500" min="0" label="Set the maximum fragment length for valid paired-end alignments" help="-X/--maxins; E.g. if `-X 100` is specified and a paired-end alignment consists of two 20-bp alignments in the proper orientation with a 60-bp gap between them, that alignment is considered valid (as long as `-I` is also satisfied).  A 61-bp gap would not be valid in that case.  If trimming options `-3` or `-5` are also used, the `-X` constraint is applied with respect to the untrimmed mates, not the trimmed mates; Deafult=500"/>
```
nsoranzo: s/Deafult/Default/

Not your fault, but while you are at it...

martenson:
Thank you very much for the PR!

nsoranzo:
Lots of good catches!

My only doubt is the option for SAM output, I think that "forcing" BAM was a way to push people to the compressed format. I'll leave that for others to comment.

Apart from that, strong +1 !

bgruening:
Great PR! Thanks a lot.
I'm with @nsoranzo. We removed this option to force the user to use a compressed format. Galaxy offers a additional tool to convert BAM to SAM if it is really needed.

peterjc:
Similarly with my ``samtools depad`` wrapper https://github.com/peterjc/pico_galaxy/tree/master/tools/samtools_depad I accept SAM or BAM input, but only offer BAM output. This also makes things like the workflow editor less fragile as the output format is fixed.

tools/bowtie2/bowtie2_wrapper.xml
```diff
@@ -147,8 +149,12 @@
...
         
         ## view/sort and output BAM file
-        | samtools view -Su - | samtools sort -o - - > $output
-        
+        #if ( str( $analysis_type.analysis_type_selector ) != "full" or str( $analysis_type.sam_opt ) != "true" ):
+          | samtools view -Su - | samtools sort -o - - > $output_1
+        #else
+          > $output_2
+        #end if
+
```
peterjc: Shouldn't these greater than signs for redirection be HTML escaped (or using CDATA)?

nsoranzo: Yes, that would be better, but the real problems come with '<' and '&'.

peterjc: Yes, those are the really problematic ones - it looks like we can get away with the greater-than-sign but it would be best practise to escape this.

@jsh58 - this was a pre-existing issue, but since you were working on this bit of the wrapper would you mind adding a commit replacing these with ``&gt;`` please?

bgruening: Actually, best practise would be to use CDATA https://wiki.galaxyproject.org/Tools/BestPractices
This can improve readability quite a lot in such cases.
If we remove the SAM options this will be removed I think, isn't it?

nsoranzo: It should go back to

```
| samtools view -Su - | samtools sort -o - - > $output
```

which still has a '>'.

jsh58:
In my workflow, it is important to have the original SAM output from Bowtie2; using the original Bowtie2 wrapper, and then converting the sorted BAM back to SAM, leaves each read's primary and secondary alignments scattered, which is useless for my purposes.

Also, it would be nice to be able to use the --reorder option in the wrapper...

nsoranzo:
I think that the fixes are important and should be merged as soon as possible, but the since the SAM output option is proving controversial, I'd suggest to split this in 2 new pull requests. Or I can merge a redacted version of your commit (maintaining you as the author), and then you can open a new PR with the SAM output change for further discussion.

Opinions?

bgruening:
@nsoranzo +1

peterjc:
Good suggestion from @nsoranzo - if @jsh58 agrees I'm +1 on merging without the SAM output option.

jsh58:
I would prefer to keep the SAM output option, for the following reasons:

1. A SAM file is the actual output from Bowtie2.

2. A sorted BAM file is not always useful, e.g. my workflow (referenced above) which requires the original SAM output.

3. For those of us with SAMTools 1.1, this avoids the Galaxy-BAM error.

nsoranzo:
@jsh58 I'm not personally against this feature, I just said that it would be better to split it from the bug fixes. May you please comment on my proposal above?

PR #57, opened 2015-01-29
Update the old sam_merge tool.

bgruening:
*  include stdio
* remove `title` field
* use best-practises
* use `multiple=True` insetead of  `<repeat>`
* increase tool version

This will hopefully address: https://biostar.usegalaxy.org/p/10441/


tools/sam_merge/sam_merge.xml
```diff
@@ -1,62 +1,56 @@
...
+            VALIDATION_STRINGENCY=LENIENT
+            O="$output1"
+            TMP_DIR=$__new_file_path__
+            #for $input in $inputs
+                I="${input}"
+            #end for
+]]>
+    </command>
+    <inputs>
+        <param name="mergeSD" value="true" type="boolean" label="Merge all component bam file headers into the merged bam file"
```
nsoranzo: Is ```value="true"``` necessary?

jj-umn: I thought boolean had attributes: truevalue=""  falsevalue=""  checked="true"

bgruening:
Nice catch @nsoranzo and @jj-umn the `value` parameter was a leftover from the very early Galaxy days as it seems. I removed it.

nsoranzo:
+1, thanks Björn!

bgruening:
Thanks @nsoranzo and @jj-umn  for reviewing!

PR #59, opened 2015-01-30
add option to output SAM from bowtie2

jsh58:
Added option to produce SAM output rather than sorted BAM.

### DIFF
--- a/tools/bowtie2/bowtie2_wrapper.xml
+++ b/tools/bowtie2/bowtie2_wrapper.xml
@@ -150,9 +150,9 @@
         
         ## output file
         #if ( str( $analysis_type.analysis_type_selector ) != "full" or str( $analysis_type.sam_opt ) != "true" ):
-          | samtools view -Su - | samtools sort -o - - > $output
+          | samtools view -Su - | samtools sort -o - - &gt; $output
         #else
-          > $output_sam
+          &gt; $output_sam
         #end if
 
         ## rename unaligned sequence files
@@ -181,12 +181,12 @@
             </param>
 
             <when value="single">
-                <param name="input_1" format="fastqsanger" type="data" label="FASTQ file" help="Nucleotide-space: Must have Sanger-scaled quality values with ASCII offset 33"/>
+                <param name="input_1" format="fastqsanger" type="data" label="FASTQ file" help="Must be of datatype &quot;fastqsanger&quot;" />
                 <param name="unaligned_file" type="boolean" truevalue="true" falsevalue="false" checked="False" label="Write unaligned reads (in fastq format) to separate file(s)" help="--un/--un-conc; This triggers --un parameter for single reads and --un-conc for paired reads" />
             </when>
             <when value="paired">
-                <param name="input_1" format="fastqsanger" type="data" label="FASTQ file" help="Nucleotide-space: Must have Sanger-scaled quality values with ASCII offset 33" />
-                <param name="input_2" format="fastqsanger" type="data" label="FASTQ file" help="Nucleotide-space: Must have Sanger-scaled quality values with ASCII offset 33" />
+                <param name="input_1" format="fastqsanger" type="data" label="FASTQ file #1" help="Must be of datatype &quot;fastqsanger&quot;" />
+                <param name="input_2" format="fastqsanger" type="data" label="FASTQ file #2" help="Must be of datatype &quot;fastqsanger&quot;" />
                 <param name="unaligned_file" type="boolean" truevalue="true" falsevalue="false" checked="False" label="Write unaligned reads (in fastq format) to separate file(s)" help="--un/--un-conc; This triggers --un parameter for single reads and --un-conc for paired reads" />
                 <conditional name="paired_options">
                     <param name="paired_options_selector" type="select" label="Do you want to set paired-end options?" help="See &quot;Alignment Options&quot; section of Help below for information">
@@ -213,7 +213,7 @@
                 </conditional>
             </when>
             <when value="paired_collection">
-                <param name="input_1" format="fastqsanger" type="data_collection" collection_type="paired" label="FASTQ Paired Dataset" help="Nucleotide-space: Must have Sanger-scaled quality values with ASCII offset 33" />
+                <param name="input_1" format="fastqsanger" type="data_collection" collection_type="paired" label="FASTQ Paired Dataset" help="Must be of datatype &quot;fastqsanger&quot;" />
                 <param name="unaligned_file" type="boolean" truevalue="true" falsevalue="false" checked="False" label="Write unaligned reads (in fastq format) to separate file(s)" help="--un/--un-conc; This triggers --un parameter for single reads and --un-conc for paired reads" />
                 <conditional name="paired_options">
                     <param name="paired_options_selector" type="select" label="Do you want to set paired-end options?" help="See &quot;Alignment Options&quot; section of Help below for information">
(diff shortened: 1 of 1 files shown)


tools/bowtie2/bowtie2_wrapper.xml
```diff
@@ -442,7 +449,29 @@
             </actions>
         </data>
         
-        <data format="bam" name="output" label="${tool.name} on ${on_string}: aligned reads (sorted BAM)">
+        <data format="bam" name="output_1" label="${tool.name} on ${on_string}: aligned reads (sorted BAM)">
```
jmchilton: Probably it would be better to still call this `output` for backward compatibility with existing workflows.

jj-umn: @jmchilton good point.   You could probably name the sam output as "output_sam".

jsh58: good suggestion, thanks

martenson:
This is a part of the original PR being split here: https://github.com/galaxyproject/tools-devteam/pull/56

BAM is preferred Galaxy format as it saves space. SAM is the original output providing more utility ( ping @jsh58 ).

Please @bgruening @nsoranzo @peterjc @blankenberg devise the best way of handling this.

peterjc:
Wasn't there some work to support unsorted BAM, or read name sorted BAM, rather than the current Galaxy default of co-ordinate sorted BAM? That would at least largely address the space concerns.

nsoranzo:
I think I'm +1 on this PR, BAM is still the default and we already expose all sort of tool options, I don't see why we should prohibit SAM output.

jgoecks:
I am +1 on this as well. In addition to Nicole's point, Bowtie2 produces SAM natively, so it makes sense that the wrapper should be able to produce SAM as well.

blankenberg:
I like the idea of allowing sam optionally, but
Should we have a single output, that changes format between bam/sam, instead of 2 different outputs?

bgruening:
I think Galaxy wrappers offer a nice opportunity to abstract complexity from the user. SAM vs. BAM vs. CRAM is complexity and we can avoid this from our users. As @peterjc already mentioned we discussed during the last GCC if we can get rid of FASTQ files. On one side to save space on the other side to abstract complexity.
I'm not sure why we should add this option, if someone really needs it there is a converter. KISS.

blankenberg:
I think there is value in having the Galaxy tool being able to provide the output from the underlying binary as created.
There is sorting occurring during the BAM creation, so it would be questionable if you could get back to the "True" Bowtie2 output.
Additionally, if you all you need is the SAM, you will end up using more space until you go back and delete the BAM file.

bgruening:
True, it's always a trade-off between features and usability. I was not aware of any tool that needs the original sorted SAM file. This is a very good point. On the other hand uses this patch `samtools sort` before writing to hard-drive.

I will not hamper this PR, please feel free to merge!

peterjc:
I was never comfortable with Galaxy's policy of automatically sorting BAM files by coordinates. There are multiple use cases for name sorted (or at least, original FASTQ read order preserving) BAM files, particularly when processing reads as pairs regardless of the mapping order.

So I'm in favour of the option to give SAM output (without applying coordinate sorting).

Regarding @blankenberg's query I can see pros and cons of having the SAM or BAM output as separate output parameters within the workflow editor context.  In some respects separate outputs probably makes life easier for linking to downstream tools (compared to a single output which changes format).

jgoecks:
Sounds like the consensus is to add it. +1 on @blankenberg's suggestion of a single output rather than multiple outputs for simplicity

martenson:
Great, thank you guys for handling this. @blankenberg is this ready to merge?

blankenberg:
I think it depends if we want to have a single output that changes format, or two outputs with one being filtered out each time.

bwlang:
I hope I'm not sticking my nose in where it's not welcome ...

I think we should not store SAM if BAM is an option. Most tools that handle SAM files can also handle BAM files - the space and I/O load savings are significant. 
 
@blankenberg makes a good point about being able to produce original output. I just did a quick test...
bowtie2 -> SAM -> BAM -> SAM
The 2 sam file are identical with the exception of the headers (which could optionally be removed) 

@peterjc I think sorting is a separate issue. One does not have to sort to produce BAM output.

I vote for BAM only, with an option to choose whether to sort (default "coordinate" for backward compatibility and to avoid having to do a distinct sort operation with the additional I/O requirements. Storing 1.5B aligned reads twice (once unsorted and once sorted) is not cool. Other choices for sort order could be "original", or "read name". I would also support burying this choice in an advanced section.
i/O and disk space are precious commodities for me, so anything we can do to encourage users not to waste them is valuable.

PS - If we do sorting by default, we should update the SO:coordinate header, which right now says unsorted for sorted bowtie2 ouptut - probably a samtools issue).

peterjc:
@bwlang I think recent samtools sort does update the SO header, but that was a long standing bug.

peterjc:
@bwlang was that test in Galaxy?

bwlang:
@peterjc - no that test was at the command line, not in galaxy
glad to hear about the SO header, maybe it's time to update my samtools ;)

peterjc:
@bwlang OK, I would have expected this to work as you described at the command line, but within Galaxy I *think* it would have added its own BAM sorting (which is ultimately why this PR exists, and why I'd like unsorted BAM to be easier in Galaxy)

jgoecks:
@blankenberg I think we want a single output for aligned reads because the tool only ever produce a single output, never two.

@bwlang Peter is correct, hence one of the reasons for supporting SAM output.

bwlang:
@jgoecks @peterjc I think SAM is unnecessary and wasteful of IO/storage also, it's unrelated to the question of order.  Why not allow specification of the order as i suggested ("coordinate", "read name" "original" - which would change the options to or eliminate the samtools sort step)?

Many tools (picard...) require coordinate sorting, so I think that makes sense as a default.  I do think it is worth including the samtools sort option to avoid the potential double disk usage of storing a sorted and unsorted file. 

BAM vs. SAM could be another choice.

I agree about the single output.

blankenberg:
The Galaxy 'bam' datatype is for a coordinate sorted BAM file. To support different (un)sorting BAM data, we should define additional datatypes.

peterjc:
Right now in the sort term, adding SAM output is the most practical way forward for this tool.

Do we all agree the more elegant medium/long term fix would be to allow the unsorted BAM as output (without Galaxy imposing co-ordinate sorting), via a new Galaxy datatype?

bwlang:
@peterjc - I agree BAM is more optimal.
Maybe I'm missing something... is there something problematic about adding an unsorted BAM datatype?

jgoecks:
The problem with an unsorted BAM datatype is that no other tool accepts it right now, so to add the datatype right now would lead to a dead end.

bwlang:
@jgoecks 
some tools work fine with unsorted BAM input (e.g. samtools idxstats), though I guess those tools would have to be modified to accept the new data type - probably not worth the effort. 

If recapitulation of the original bowtie output is not the default (BAM) it probably won't be frequently used and probably doesn't matter very much whether it's unsorted BAM or SAM.

Is it inconsistent that the galaxy SAM datatype does not specify a sort order, but galaxy's BAM data type is implicitly coordinate sorted? (I was not aware of the BAM coordinate sort assumption until this discussion.)

peterjc:
@bwlang @jgoecks That may have been a typo, but ``samtools idxstats`` needs an indexed BAM file, which in turn means the BAM file must be coordinate sorted. So not a good example.

Meanwhile in the short term, are there any objections to @jsh58's short term solution of adding unsorted SAM output as an option for the bowtie2 wrapper?

nsoranzo:
@peterjc I think the majority is in favour of adding the optional SAM output, the last remaining doubt  (raised by @blankenberg) is if it should be a single output.

jgoecks:
Single output: it's never the case the tool will produce both a BAM and SAM output. The only question I have is whether a multi-type, parameter dependent output will work in the workflow editor.

bgruening:
@jsh58 do you mind changing the outputs to either SAM *or* BAM?

jsh58:
As written, it produces either output.

bgruening:
So I think @bwlang and me are in a minority here, so I willing to merge this and resolve the merge conflict.
Any last minute concerns? Ping @peterjc @blankenberg @nsoranzo

martenson:
related discussion here: https://twitter.com/KeywanHP/status/646661366774923264

bgruening:
This one was merged as rebase in https://github.com/galaxyproject/tools-devteam/pull/230

PR #62, opened 2015-02-19
Add a Data Manager that will communicate with Galaxy Project's rsync ser...

blankenberg:
...ver as the source for the reference Data. Data Tables not already defined on the host will be dynamically loaded from Galaxy Main's configuration if the admin requests data for the missing tool data table.


martenson:
:+1:

PR #64, opened 2015-02-24
Velvet (+package) from galaxy

erasche:
Imported the commented out sr_assembly tool velvet, added a package for it.

*it would be nice* if Galaxy had configurable package installs as part of the future. It's, admittedly, a bit of a corner case, but velvet has the setting `MAXKMERLENGTH` which we need set higher than the default. Instead I just re-built the tool and copied the binaries manually into the install dir.

### DIFF
--- a/packages/package_velvet_1_2_10/tool_dependencies.xml
+++ b/packages/package_velvet_1_2_10/tool_dependencies.xml
@@ -1,0 +1,31 @@
+<tool_dependency>
+    <package name="velvet" version="1.2.10">
+        <install version="1.0">
+            <actions>
+                <action type="download_by_url">https://github.com/dzerbino/velvet/archive/9adf09f7ded7fedaf6b0e5e4edf9f46602e263d3.zip</action>
+                <action type="shell_command">make</action>
+                <action type="move_file">
+                    <source>velvetg</source>
+                    <destination>$INSTALL_DIR</destination>
+                </action>
+                <action type="move_file">
+                    <source>velveth</source>
+                    <destination>$INSTALL_DIR</destination>
+                </action>
+                <action type="set_environment">
+                     <environment_variable name="PATH" action="prepend_to">$INSTALL_DIR</environment_variable>
+                 </action>
+            </actions>
+        </install>
+        <readme>
+From the official readme:
+
+::
+
+    Velvet should function on any standard 64bit Linux environment with
+    gcc. A good amount of physical memory (12GB to start with, more is no luxury)
+    is recommended.
+
+        </readme>
+    </package>
+</tool_dependency>
--- a/packages/package_velvet_1_2_10/tool_dependencies.xml
+++ b/packages/package_velvet_1_2_10/tool_dependencies.xml
@@ -2,7 +2,7 @@
     <package name="velvet" version="1.2.10">
         <install version="1.0">
             <actions>
-                <action type="download_by_url">https://github.com/dzerbino/velvet/archive/9adf09f7ded7fedaf6b0e5e4edf9f46602e263d3.zip</action>
+                <action type="download_by_url">https://github.com/dzerbino/velvet/archive/d430f1ade319af582dc7bb6886ea904abd6337bd.zip</action>
                 <action type="shell_command">make</action>
                 <action type="move_file">
                     <source>velvetg</source>
--- a/tools/velvet/velvetg.xml
+++ b/tools/velvet/velvetg.xml
@@ -1,0 +1,301 @@
+<tool id="velvetg" name="velvetg" version="1.0.0">
+  <description>Velvet sequence assembler for very short reads</description>
+  <version_command>velvetg 2&gt;&amp;1 | grep "Version" | sed -e 's/Version //'</version_command>
+  <command interpreter="python">
+    velvetg_wrapper.py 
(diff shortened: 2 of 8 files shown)


packages/package_velvet_1_2_10/tool_dependencies.xml
```diff
@@ -0,0 +1,31 @@
+<tool_dependency>
+    <package name="velvet" version="1.2.10">
+        <install version="1.0">
+            <actions>
+                <action type="download_by_url">https://github.com/dzerbino/velvet/archive/9adf09f7ded7fedaf6b0e5e4edf9f46602e263d3.zip</action>
```
nsoranzo: Why not http://www.ebi.ac.uk/~zerbino/velvet/velvet_1.2.10.tgz ?

bgruening: Github is more trustable than a home directory?

tools/velvet/velvetg.xml
```diff
@@ -0,0 +1,310 @@
...
+        <param name="ins_length" value = "-1" label="Insert Length in Paired-End Read dataset (ignored when -1)" type="integer" help="Expected distance between two paired end reads"/>
+        <conditional name="options">
+          <param name="advanced" type="select" label="Velvet Advanced Options">
+            <option value="no">Use Defaults</option>
+            <option value="yes">Set Advanced Option Values</option>
+          </param>
+          <when value="no"/>
+          <when value="yes">
+            <param name="ins_length_sd" value = "-1" label="Estimate of Standard Deviation of Paired-End Read dataset(ignored when -1)" type="integer" help="Estimate of standard deviation of Paired-End Read dataset (default: 10% of corresponding length)"/>
+            <param name="ins_length2" value = "-1" label="Insert Length in 2nd Paired-End Short Read dataset (ignored when -1)" type="integer" help="Expected distance between two paired end reads in the second short-read dataset"/>
```
bgruening: The spaces between `value` and `=` should be removed.

tools/velvet/velvetg_wrapper.py
```diff
@@ -0,0 +1,46 @@
+#!/usr/bin/env python
+
+"""
+Classes encapsulating decypher tool.
+James E Johnson - University of Minnesota
+"""
+import os
```
bgruening: I think we can get rid of this python file and use the `<stdio>` tag.

tools/velvet/velvetg.xml
```diff
@@ -0,0 +1,310 @@
+<tool id="velvetg" name="velvetg" version="1.0.0">
+  <description>Velvet sequence assembler for very short reads</description>
+  <version_command>velvetg 2&gt;&amp;1 | grep "Version" | sed -e 's/Version //'</version_command>
+  <requirements>
+      <requirement type="package" version="1.2.10">velvet</requirement>
+  </requirements>
+  <stdio>
+      <exit_code range="1:" />
```
bgruening: `requirements` and `stdio` should move into a macro file.

packages/package_velvet_1_2_10/tool_dependencies.xml
nsoranzo: Yes, but the releases are not tagged on GitHub.

I think that @erasche just took the last commit id, while the 1.2.10 release corresponds to https://github.com/dzerbino/velvet/commit/d430f1ade319af582dc7bb6886ea904abd6337bd (had to find it manually...).

erasche: That's exactly what I did. Personal opinion but as per @bgruening's comment, I trust an SHA1 sum from a git repo a lot more than a mutable file in someone's homedir.

tools/velvet/velvetg.xml
erasche: can do.

packages/package_velvet_1_2_10/tool_dependencies.xml
nsoranzo: Then I'd prefer this to be changed to d430f1ade319af582dc7bb6886ea904abd6337bd.zip or change the package version to the commit ID.

erasche: @nsoranzo is that the commit that corresponds to what can be downloaded from their website?

Otherwise, https://github.com/dzerbino/velvet/commit/5cea317b4ec5c3f0fd2931321bc7231d8a250682 is a good option since that's when the version bump occurred.

nsoranzo: Yes, it is exactly that, as I mentioned above.

nsoranzo:
Yeah, I looked at the code in https://github.com/galaxyproject/galaxy/blob/dev/lib/tool_shed/galaxy_install/tool_dependencies/recipe/step_handler.py and it looks safe.

+1 if you correct the zip to the correct id.

tools/velvet/velvetg_wrapper.py
erasche: probably, but that's a lot of work, and outside the scope of what I can reasonably justify right now :(

erasche:
Code review comments addressed.

bgruening:
+1 we should get this in and fix the `help` and the additional python script later.

PR #68, opened 2015-02-25
remove additional inputs from the user interface

bgruening:
This can be done with `multiple=True`.

### DIFF
--- a/tool_collections/cufflinks/cuff_macros.xml
+++ b/tool_collections/cufflinks/cuff_macros.xml
@@ -72,19 +72,11 @@
   </token>
   <xml name="cufflinks_gtf_inputs">
     <param format="gtf" name="inputs" type="data" label="GTF file(s) produced by Cufflinks" help="" multiple="true" />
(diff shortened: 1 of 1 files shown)


bgruening:
Thanks to Pavan for spotting this.

jmchilton:
It is this way intentionally - for now you can only substitute a single collection into the ``multiple=True`` data input - but you may wish to supply multiple collections to this tool - right? That is why that repeat is there and the input type only allows collections.

I'll admit this is confusing - maybe we should update the help on the repeat input instead.

bgruening:
John what happens if you add more than one collection? Would it be the same result as putting all files into one collection instead of two?

martenson:
ping @jmchilton - please have a look

jmchilton:
Yes @bgruening - if you look at the for loop - they are all just sort of merged together in one big list of files to pass into the tool.

martenson:
Thank you for the clarification. Closing now.

PR #71, opened 2015-02-26
This is not strictly needed, but we are not sure why this is working.

bgruening:
It should not work, at least all other parameters behave differently. See also https://github.com/galaxyproject/galaxy/commit/cf8fd5c5970b0bc5f6dd4f8a9af710a29d90f2e4.


nsoranzo:
+1

PR #73, opened 2015-03-09
Make tool version explicit (Tool Shed would assume 1.0.0)

peterjc:
See discussion on IUC mailing list, CC @davebx


davebx:
Loud :+1: from me.

peterjc:
Pushed an update with update commit message (Tool Shed --> Job runner)

PR #77, opened 2015-03-16
Remove line continuation char (\) from GATK Picard Index Data Manager

blankenberg:
NB: We should determine if this used to work as-is, and if some job running change in Galaxy broke this tool.

Example of behavior, notice that the arguments after a \ have a space prepended before the '--argname':

argv ['/path/galaxy-blankenberg/installed/toolshed.g2.bx.psu.edu/repos/devteam/data_manager_gatk_picard_index_builder/a15709ad26ea/data_manager_gatk_picard_index_builder/data_manager/data_manager_gatk_picard_index_builder.py', '/path/galaxy-blankenberg/database/files/000/dataset_4.dat', ' --jar', '/path/shed_tools/galaxy-blankenberg/tool_dependency_dir/picard/1.56.0/devteam/package_picard_1_56_0/99a28567c3a3/jars/CreateSequenceDictionary.jar', ' --fasta_filename', '/path/galaxy-blankenberg/tool-data/hg19/seq/hg19.fa', ' --fasta_dbkey', 'hg19', ' --fasta_description', 'hg19', ' --data_table_name', 'gatk_picard_indexes']

### DIFF
--- a/data_managers/data_manager_gatk_picard_index_builder/data_manager/data_manager_gatk_picard_index_builder.xml
+++ b/data_managers/data_manager_gatk_picard_index_builder/data_manager/data_manager_gatk_picard_index_builder.xml
@@ -5,11 +5,11 @@
         <requirement type="package" version="1.56.0">picard</requirement>
     </requirements>
     <command interpreter="python">
-        data_manager_gatk_picard_index_builder.py "${out_file}" \
-            --jar "\$JAVA_JAR_PATH/CreateSequenceDictionary.jar" \
-            --fasta_filename "${all_fasta_source.fields.path}" \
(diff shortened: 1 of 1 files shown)


blankenberg:
@davebx Did you add this DM? Did it use to work correctly?

@jmchilton (or anyone else) any thoughts on changes in the last year that might have caused a change in behavior as described here.

blankenberg:
It is possible that this tool may have only worked previously on the local runner, which used to run commands differently than the other runner plugins, but was changed recently to work in the same fashion. 
After local discussion, we're just going to merge this and recommend not using line continuation characters in a <command> tag.
If anyone else has additional info/comments, please feel free to keep the discussion going.

PR #76, opened 2015-03-13
adopt Anton's vcflib tools

martenson:

PR #78, opened 2015-03-16
vcflib tools adoption

martenson:
same PR as https://github.com/galaxyproject/tools-devteam/pull/76 but in its own branch so people can actually use it


jmchilton:
Thanks for moving the branch to tools-devteam, I am going to take a crack at introducing a macro file with citations.

jmchilton:
Some of these tools use a `stdio` tag and some don't. Any problem with me just reworking all of them to use the same stdio tag (error on exit code != 0). Ping @nekrut.

martenson:
@jmchilton there is no problem with that John, commit ahead!

PR #79, opened 2015-03-16
bump samtools to v1.2 and use macros

martenson:
also create a symlink for tool_dependencies.xml

### DIFF
--- a/tool_collections/samtools/samtools_bedcov/samtools_bedcov.xml
+++ b/tool_collections/samtools/samtools_bedcov/samtools_bedcov.xml
@@ -1,8 +1,9 @@
-<tool id="samtools_bedcov" name="Calculate read depth" version="1.0.0">
-    <description>on BAM files</description>
-    <requirements>
-        <requirement type="package" version="1.1">samtools</requirement>
-    </requirements>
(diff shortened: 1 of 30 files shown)


tool_collections/samtools/samtools_bedcov/samtools_bedcov.xml
```diff
@@ -1,8 +1,9 @@
...
-        <requirement type="package" version="1.1">samtools</requirement>
-    </requirements>
+<tool id="samtools_bedcov" name="Calculate read depth" version="1.0.1">
+  <description>on BAM files</description>
+  <macros>
+    <import>macros.xml</import>
+  </macros>
+  <expand macro="requirements"></expand>
     <version_command>samtools --version | head -n 1 | awk '{ print $2 }'</version_command>
     <command><![CDATA[
```
bgruening: I think you can also move the version_command into the macro. `sam_to_bam` is missing the version_command for example.

martenson: done, good idea!

bgruening:
Thanks @martenson! :+1:

PR #80, opened 2015-03-16
Fix cummeRbund and cummerbund_to_tabular

davebx:


bgruening:
Do we need to depend now on sqlite? I think we have a repository in TTS.

davebx:
If I'm not mistaken, it's built-in since 2.5.

jmchilton:
https://docs.python.org/2/library/sqlite3.html

:+1:  Yes - this is excellent! We should not be importing galaxy stuff - it means we can never refactor Galaxy and also means we cannot reuse the tool shed artifacts outside of Galaxy. I like this a lot better.

bgruening:
Very nice!  :+1:

PR #86, opened 2015-03-19
Fix Bowtie2 tool not failing properly when non-zero exit code is provide...

blankenberg:
...d. We cannot use pipes due to needing the return code from the initial command and we can't assume that we are bash or another shell that has e.g. PIPESTATUS.

### DIFF
--- a/tools/bowtie2/bowtie2_wrapper.xml
+++ b/tools/bowtie2/bowtie2_wrapper.xml
@@ -154,10 +154,10 @@
 
         ## rename unaligned sequence files
         #if $library.type == "paired" and $output_unaligned_reads_l and $output_unaligned_reads_r:
-            #set left  = str($output_unaligned_reads_l)[:-3] + '1.dat'
-            #set right = str($output_unaligned_reads_l)[:-3] + '2.dat'
-            &amp;&amp; mv "${ left }" "${ output_unaligned_reads_l }"
-            &amp;&amp; mv "${ right }" "${ output_unaligned_reads_r }"
+            #from os.path import splitext
+            #set _unaligned_root, _unaligned_ext = splitext( str( $output_unaligned_reads_l ) )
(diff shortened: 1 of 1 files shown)


tools/bowtie2/bowtie2_wrapper.xml
```diff
@@ -145,18 +145,19 @@
...
 
         ## rename unaligned sequence files
         #if $library.type == "paired" and $output_unaligned_reads_l and $output_unaligned_reads_r:
-            #set left  = str($output_unaligned_reads_l).replace( '.dat', '.1.dat' )
-            #set right = str($output_unaligned_reads_l).replace( '.dat', '.2.dat' )
-        
-            ; mv $left $output_unaligned_reads_l;
-            mv $right $output_unaligned_reads_r
+            #set left  = str($output_unaligned_reads_l)[:-3] + '1.dat'
+            #set right = str($output_unaligned_reads_l)[:-3] + '2.dat'
```
bgruening: I haven't tested it but the `1.dat` looks strange to me. Is bowtie2 creating `*.dat` files?

blankenberg: Ah, yes, you are probably correct here, what we really need a replace from right, count of 1 (directory could have .dat as part of the path). I'll update this to fix it.

blankenberg: This looks ok. Bowtie must be spliting on the ext and then adding the 1/2:
-rw-r--r--    1 dan  staff   178B Mar 20 08:46 dataset_186.1.dat
-rw-r--r--    1 dan  staff   178B Mar 20 08:46 dataset_186.2.dat
-rw-r--r--    1 dan  staff     0B Mar 20 08:46 dataset_186.dat

It likely looks less odd when the passed in output filename is e.g., unaligned.fa.

I don't like that it works on the assumption that the output files all end in .dat (we should probably do a splitext to figure it out), but that isn't a change in behavior.

PR #91, opened 2015-03-23
Update mpileup outputs

davebx:


tool_collections/samtools/samtools_mpileup/samtools_mpileup.xml
```diff
@@ -73,98 +68,97 @@
...
+                        <option value="--VCF">VCF</option>
+                        <option value="--BCF">BCF</option>
+                    </param>
+                    <when value="--VCF">
+                        <param checked="False" falsevalue="--uncompressed" label="Compress output" name="compressed" truevalue="" type="boolean" />
+                    </when>
+                    <when value="--BCF">
+                        <param checked="False" falsevalue="--uncompressed" label="Compress output" name="compressed" truevalue="" type="boolean" />
+                    </when>
+                </conditional>
```
nsoranzo: I think this `<conditional>` block can be simplified to:

```
<param label="Choose the output format" name="output_format_selector" type="select">
    <option value="--VCF">VCF</option>
    <option value="--BCF">BCF</option>
</param>
<param checked="False" falsevalue="--uncompressed" label="Compress output" name="compressed" truevalue="" type="boolean" />
```

PR #96, opened 2015-04-08
Update cufflinks_wrapper.py

bgruening:
Without this change we will get the following error:
```
Usage: cufflinks_wrapper.py [options]
cufflinks_wrapper.py: error: --compatible-hits-norm option requires an argument
```


bgruening:
Do we need to increasing the version number? I think it was not able to use this option. So actually it should not affect reproducibility.

nsoranzo:
:+1: 

No need to increase version number IMO.

bgruening:
@nsoranzo thanks for reviewing!
@martenson can you please update this? It's an important bugfix.

martenson:
Repository tools-devteam/tool_collections/cufflinks/cufflinks updated successfully.

martenson:
updated on test.galaxyproject.org too

PR #98, opened 2015-04-09
Samtools update

nekrut:
Updates to most samtools.


tool_collections/samtools/samtools_stats/samtools_stats.xml
```diff
@@ -1,8 +1,12 @@
-<tool id="samtools_stats" name="Generate statistics" version="1.0.1">
-    <description>for a BAM or SAM file</description>
+<tool id="samtools_stats" name="Stats" version="1.0.1">
```
martenson: I think version needs to be bumped.

tool_collections/samtools/samtools_reheader/samtools_reheader.xml
```diff
@@ -1,5 +1,5 @@
-<tool id="samtools_reheader" name="Replace header" version="1.0.1">
-    <description>on a BAM or SAM file</description>
+<tool id="samtools_reheader" name="Reheader" version="1.2.1">
```
martenson: Does the version here need to be raised?

tool_collections/samtools/samtools_bedcov/samtools_bedcov.xml
```diff
@@ -1,5 +1,5 @@
-<tool id="samtools_bedcov" name="Calculate read depth" version="1.0.1">
-  <description>on BAM files</description>
+<tool id="samtools_bedcov" name="BedCov" version="1.0.1">
+  <description>calulate read depth for a set of genomic intervals</description>
```
nsoranzo: Typo: s/calulate/calculate/

nekrut: tx!

tool_collections/samtools/samtools_reheader/samtools_reheader.xml
nekrut: yep

nekrut:
Let's me make a few more changed. So do not merge yet!

tool_collections/samtools/samtools_reheader/samtools_reheader.xml
martenson: I see only name, description, param labels, tests and help changes. And one more format for input_header. 
Is the additional format why it needs a new version?

nekrut:
Please, review now. Txs to Dan and DaveB for tweaks.

tool_collections/samtools/sam_to_bam/sam_to_bam.xml
```diff
@@ -1,5 +1,5 @@
-<tool id="sam_to_bam" name="SAM-to-BAM" version="1.1.6">
-  <description>converts SAM format to BAM format</description>
+<tool id="sam_to_bam" name="SAM-to-BAM" version="1.1">
```
nsoranzo: Here the version number is decreased (on main Tool Shed latest version is 1.1.4).

martenson: good catch, it was 1.1.6 = should be 1.1.7 at least now

nekrut:
@martenson: I wanted to sync all samtools to a single release as before if was a mess. however, I can obviously change those back. To me once we release the new set with the same base version - it will be better. No?

martenson:
I think users mostly don't care about the Galaxy tool version numbers being unified within the tool collection and this forces deployers to do more new installations, forces new suites to be created and creates a precedent. What are you going to do when one of the Samtools tools will need an update next time? I think we did split tools into separate repositories so we can avoid these version increments.

nsoranzo:
+1 to increase only when necessary.

nekrut:
I solemnly swear to only increase versions only as necessary in all future releases. I would like to keep all samtools at 1.1 for this one as this is the fists standardized release for our samtools collection.

nsoranzo:
Ok for me, but sam_to_bam should be at 1.1.7 . Decreasing versions is a no-go.

nekrut:
ok

tool_collections/samtools/sam_to_bam/sam_to_bam.xml
```diff
@@ -16,14 +16,14 @@
             ln -s ${source.index.fields.path} input.fa &&
             ln -s ${source.index.fields.path}.fai input.fa.fai &&
           #end if
-          samtools view -T input.fa sorted_input.bam -b -h -o $output1
+          samtools view -@ \${GALAXY_SLOTS:-1} -b -h -o $output1 -T input.fa sorted_input.bam
```
bgruening: Oh yes, parallel samtools! Thanks @nekrut !

tool_collections/samtools/samtools_stats/samtools_stats.xml
```diff
@@ -1,8 +1,12 @@
...
-    <description>for a BAM or SAM file</description>
+<tool id="samtools_stats" name="Stats" version="1.1">
+    <description>generate statistics for BAM dataset</description>
     <macros>
     <import>macros.xml</import>
   </macros>
+    
+    <!-- the code file below should be enabled with samtools -I bug is fixed -->
+    <!--<code file="samtools_stats_options.py"/>-->
+
```
bgruening: @nekrut can you provide some more infos here? If we depend on a strict samtools version, is this needed or not?

tool_collections/samtools/samtools_stats/samtools_stats_options.py
```diff
@@ -0,0 +1,44 @@
+import pysam
+import os
+import tempfile
```
bgruening: This file should get some PEP8 love and probably a shebang.

bgruening:
I like this PR!

tool_collections/samtools/samtools_stats/samtools_stats_options.py
blankenberg: Hopefully can actually get rid of these code files and use the newly extended BAM metadata :)

tool_collections/samtools/samtools_stats/samtools_stats.xml
nekrut: looks like it is broken in all versions

tool_collections/samtools/samtools_stats/samtools_stats_options.py
nekrut: This will go away completely. Instead I will use rely on BAM metadata, which now includes refs, lengths, RGs etc. thanks to a recent update by @blankenberg

nekrut:
I will add requested changes in the next few days. DO NOT MERGE

nekrut:
Please, complain and/or merge

bgruening:
LGTM, if no one else complains I will merge after dinner :)

PR #102, opened 2015-04-13
fix installation where python will install libraries into `$HOME/lib64`

bgruening:
This ensures that libraries are installed into `$HOME/lib64`.


martenson:
Thanks @bgruening

PR #104, opened 2015-04-14
Add data manager to generate kraken databases.

davebx:
Ping @blankenberg for obvious reasons.

### DIFF
--- a/data_managers/data_manager_build_kraken_database/data_manager/kraken_database_builder.xml
+++ b/data_managers/data_manager_build_kraken_database/data_manager/kraken_database_builder.xml
@@ -1,5 +1,5 @@
 <?xml version="1.0"?>
-<tool id="kraken_database_builder" version="1.0.0" tool_type="manage_data" name="Kraken">
+<tool id="kraken_database_builder" name="Kraken" tool_type="manage_data" version="1.0.0">
     <description>database builder</description>
     <requirements>
         <requirement type="package" version="0.10.5">kraken</requirement>
@@ -7,7 +7,7 @@
         <requirement type="set_environment">KRAKEN_SCRIPT_PATH</requirement>
     </requirements>
     <stdio>
-        <exit_code range="1:" level="fatal" description="Error" />
+        <exit_code description="Error" level="fatal" range="1:" />
     </stdio>
     <version_command>kraken -version | awk '{print $NF}'</version_command>
     <command>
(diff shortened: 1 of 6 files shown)


data_managers/data_manager_build_kraken_database/tool_dependencies.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0"?>
+<tool_dependency>
+    <set_environment version="1.0">
+        <environment_variable action="set_to" name="KRAKEN_SCRIPT_PATH">$REPOSITORY_INSTALL_DIR</environment_variable>
```
jmchilton: ``_SCRIPT_PATH`` is a deprecated idiom - I would eliminate this and just use ``$__tool_directory__`` in the tool.

For more information - see https://bitbucket.org/galaxy/galaxy-central/commits/ddbe0d12bc86da406bb96213e23b8ddb273ed669 and bgruening/galaxytools#85.

data_managers/data_manager_build_kraken_database/data_manager/kraken_database_builder.xml
```diff
@@ -0,0 +1,45 @@
...
+    <inputs>
+        <param name="database_name" type="text" label="Name for this database" />
+        <param name="libraries" type="select" multiple="True" label="Select partial library to download">
+            <option value="bacteria">Bacteria</option>
+            <option value="plasmids">Plasmids</option>
+            <option value="viruses">Viruses</option>
+            <option value="human">Human</option>
+        </param>
+        <repeat name="fasta_add" label="FASTA File" title="Add FASTA files to the library">
+            <param name="input_fasta" type="data" format="fasta" label="Select history item" />
```
bgruening: I think you can remove the repeat tag here and simple use `multiple="True"` in the data parameter.

data_managers/data_manager_build_kraken_database/tool_dependencies.xml
blankenberg: This is a good thing going forward, but I do seriously wonder if it is worth needlessly increasing the minimal Galaxy version that would then be required to use this tool. 

e.g. are any other newer/same aged features used?

jmchilton: I would also not support *needlessly* increasing the minimal Galaxy version (to a point) but tools that use this idiom are pretty much broken if not installed via the tool shed - there are potentially workarounds sure - but I think making the tools easy to use without going through the install process is very important.

You get to pick your kind of broken - broken pre-15.03 or broken if not installed via the Tool Shed forever. I much prefer the first kind of broken.

erasche: Sounds like Galaxy needs to start having a `minimum_version` tag on tool and warn/error if installed tools don't meet that requirement.

bgruening: @jmchilton if anyone is not using the TS and installing by it's own, he is probably smart enough to do the `env.sh` dance :)
But yes I'm always in favour of removing old cruft.

bgruening:
I think we can merge this!?

data_managers/data_manager_build_kraken_database/tool_dependencies.xml
jmchilton: A bug having a workaround doesn't mean the bug doesn't exist :) - that said lets merge this.

Issue #106, opened 2015-04-16
cummeRbund is missing .shed.yml file


nsoranzo:
There is no `.shed.yml` file in `tools/cummeRbund/`. I would add one, but I'm not sure if we want to (and can) reuse jjohnson's Tool Shed repository.

PR #107, opened 2015-04-16
Add .shed.yml file for cummeRbund

bgruening:
fix https://github.com/galaxyproject/tools-devteam/issues/106


nsoranzo:
So we go with the new TS repository?

jj-umn:
I think so.  That would give the tool the devteam branding.

bgruening:
Anyone? @nsoranzo?

nsoranzo:
:+1: 

That was my only concern.

PR #124, opened 2015-04-28
Create cd-hit-dup tool that uses new cd-hit-auxtools package.

blankenberg:


martenson:
Are you sure you want to add the package here (and not to IUC)? (and also to name it `_fix_dan` ?)

blankenberg:
We can put the package and/or tool under iuc, doesn't matter to me really.

We can drop the _dan from the name, but the _fix should probably stay. I made a GitHub repo from ```cd-hit-auxtools-v0.5-2012-03-07-fix.tar.gz```, with some minor modifications.

tools/cd_hit_dup/cd_hit_dup.xml
```diff
@@ -0,0 +1,135 @@
...
+            <when value="single">
+                <param name="fastq_input1" type="data" format="fastqsanger,fasta" label="Select read dataset" help="Specify dataset with single reads"/>
+                <conditional name="filter_chimeras">
+                    <param name="filter_chimeras_selector" type="select" label="Filter out chimeric clusters">
+                        <option value="true">Yes</option>
+                        <option value="false" selected="True">No</option>
+                    </param>
+                    <when value="true">
+                        <param name="min_chimeric_length" type="integer" value="30" min="20" label="Minimum length of common sequence shared between a chimeric read and each of its parents" help="-s"/>
+                        <param name="abundance_cutoff" type="integer" value="1" min="1" label="Abundance cutoff" help="-a; Tool Author recommend default of 2, but this would require the chimera itself to need 2 copies"/>
```
blankenberg: Any comments/concerns for "Tool Author recommend default of 2, but this would require the chimera itself to need 2 copies", since I set the default to 1, not the program's recommended 2.

bgruening: I'm ok with this. As alternative we could set the default to 2 and recommend 1.

martenson:
Name makes sense. I think the package would be better in IUC repo though, would be great if you could retarget it. I believe the consensus is that all packages belong to the IUC repo.

blankenberg:
I'll push the package to IUC.

erasche:
We'll look forward to it! :)

blankenberg:
Would like to get live on Test soon. Any outstanding issues? ping @martenson

martenson:
deploying now

PR #129, opened 2015-05-08
Correct reStructuredText typo

Unode:
The dashed line is being treated as marker for a section header instead of as an horizontal line.


bgruening:
Thanks @Unode! This looks good to me.

PR #131, opened 2015-05-08
Silence warnings due to missing <when> tags

Unode:
Two tools were missing tags causing warnings when parsed by galaxy.


blankenberg:
Thanks!

PR #134, opened 2015-05-18
introduce suite for new cufflinks

martenson:


suites/suite_cufflinks_2_2_1/.shed.yml
```diff
@@ -0,0 +1,2 @@
+owner: devteam
```
bgruening: @martenson this should have more informations, right? Would it be worth to have an own category for suites?

erasche: all of the devteam suites are like this. I was thinking/hoping we wouldn't even need suite folders if planemo was going to handle the decomposition...

martenson: I think suites are mostly for the easiness of installation into Galaxy more than anything else. I agree it could have more information but devteam repo did not adopt the IUC advanced techniques yet.

PR #135, opened 2015-05-20
Minor updates to kraken.

davebx:
Set the kraken DB path in an environment variable.
Remove the histogram output definition.
Fix for when no external FASTA files are defined.

### DIFF
--- a/data_managers/data_manager_build_kraken_database/data_manager/kraken_database_builder.xml
+++ b/data_managers/data_manager_build_kraken_database/data_manager/kraken_database_builder.xml
@@ -21,7 +21,9 @@
             #end if
         #end for
         #for $fasta in $input_fasta:
-            kraken-build --threads \${GALAXY_SLOTS:-1} --db "${database_name}" --add-to-library "$fasta"  &&
+            #if str($fasta) != 'None':
+                kraken-build --threads \${GALAXY_SLOTS:-1} --db "${database_name}" --add-to-library "$fasta"  &&
+            #end if
         #end for
         kraken-build --threads \${GALAXY_SLOTS:-1} --db "${database_name}" --build --kmer-len ${kmer_len} --minimizer-len ${minimizer_len}  &&
         cd .. && python \$KRAKEN_SCRIPT_PATH/data_manager/make_json.py --db "${database_name}" --out "${out_file}"
--- a/tool_collections/kraken/kraken/kraken.xml
+++ b/tool_collections/kraken/kraken/kraken.xml
@@ -19,9 +19,9 @@
 
         ${only_classified_output}
 
-        #if str( $quick-operation.quick ) == "yes":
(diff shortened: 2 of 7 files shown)


bgruening:
If the select is `optional="False"` this should not happen afaik and `optional="False"` is the default setting if not given, isn't?

nsoranzo:
Moreover, even if `$input_fasta` is empty, `$fasta` cannot be `None`.

davebx:
I just checked locally. When no input fasta is provided, `$input_fasta` contains: `['None']`

nsoranzo:
This sounds like a Galaxy bug, I would expect `[]`.

bgruening:
@davebx I would consider this a Galaxy bug.
Can you add an explicit "optional='True'" to this argument?

nsoranzo:
@bgruening You mean `optional="false"`, don't you?

davebx:
The input fasta option should actually be optional, kraken-build doesn't *require* providing additional sequences for building its libraries.

bgruening:
@nsoranzo I guess it's either a `database` or a `FASTA file` or both, as far as I can see. 
@davebx knows this  it better.

blankenberg:
In the very least, that 'None' should actually be a NoneDataset object that has a boolean value of False (via __nonzero__)  and a str() of "None" (as is the case for a single optional dataset).

But, the probable behavior of a multiple dataset that is optional with no option selected should be an empty iterable with boolean value of False (via __nonzero__ and __len__=0), that when the iterable is directly hit with str() should result in "None".

nsoranzo:
@blankenberg I suggested a simple `[]`, but an empty iterable works as well :)

blankenberg:
In a tool command tag/config file, etc:

Is str(multiple_dataset) output as ```'["path1", "path2"]'```, where str(single_dataset) is ```"path1"```?
Whereas str( multiple_select ) is output as ```'value1,value2'```, with str(single_select) as ```"value1"```? 

if so, yikes.

bgruening:
@blankenberg seems to yes :(

blankenberg:
Well thats not good, I'd consider ```str(multiple_dataset) output as '["path1", "path2"]'``` to be a bug, but I wonder if any tool has actually been designed to anticipate this behavior.

davebx:
Rebased #148

bgruening:
@blankenberg do we want to fill this as bug? Can we grep in the TS to see if this was already used?

blankenberg:
@bgruening Yes, we should do both, although a bit more complicated to grep for a tool expecting a string representation of a list.

bgruening:
@blankenberg can we merge this? The `bug` is on your ToDo list right? ;)

blankenberg:
I am +1 with the merge.

Made a trello card: https://trello.com/c/WPxFNRqq for this issue, although I am happy if someone else wants to get to it before I have a chance.

PR #137, opened 2015-05-20
Add HISAT wrapper and data manager.

davebx:


data_managers/data_manager_hisat_index_builder/tool-data/hisat_indexes.loc.sample
```diff
@@ -0,0 +1,37 @@
...
+# The path points to the "basename" for the set, not a specific file.
+# It has four text columns seperated by TABS.
+#
+# <unique_build_id>	<dbkey>	<display_name>	<file_base_path>
+#
+# So, for example, if you had hg18 indexes stored in:
+#
+#    /depot/data2/galaxy/hg19/hisat/
+#
+# containing hg19 genome and hg19.*.bt2 files, such as:
```
bgruening: Are the new indices also called bt2?

tools/hisat/hisat.xml
```diff
@@ -0,0 +1,344 @@
+<tool id="hisat" name="HISAT" version="1.0.0">
+    <description>spliced aligner</description>
+    <macros>
+        <import>hisat_macros.xml</import>
+    </macros>
+    <requirements>
+        <requirement type="package" version="0.1.5">hisat</requirement>
+        <requirement>samtools</requirement>
```
bgruening: Do we want to depend on a specific samtools verison?

tools/hisat/tool_dependencies.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0"?>
+<tool_dependency>
```
bgruening: samtools?

tools/hisat/hisat.xml
```diff
@@ -0,0 +1,344 @@
...
+        #end if
+        #if str($spliced_options.spliced_options_selector) == "advanced":
+            --pen-cansplice ${spliced_options.canonical_penalty} --pen-noncansplice ${spliced_options.noncanonical_penalty}
+            --pen-intronlen ${spliced_options.function_type},${spliced_options.constant_term},${spliced_options.coefficient_term}
+            #if str($spliced_options.known_splice_gff) != 'None':
+                --known-splicesite-infile splice_sites.txt
+            #end if
+        #end if
+        -S alignments.sam &&
+        samtools view -bS alignments.sam -o alignments.bam &&
```
bgruening: Can we improve this to save IO, using a pipe and the `-o` option? Moreover, if we require some recent version of samtools we can use multiple threads.

bgruening:
@davebx really looking forward to get this tool to our server! Thanks for working on this! 
Besides the samtools requirements from me a :+1:

data_managers/data_manager_hisat_index_builder/tool-data/hisat_indexes.loc.sample
davebx: So it's similar to bowtie2's indexes, but trying to use bowtie2-build generated indexes was a massive failure.

tools/hisat/hisat.xml
davebx: Possibly, now that you mention it. I'll update it to depend on either 0.1.18 or 0.1.19 by IUC.

bgruening:
@davebx any reason to not use the latest samtools version?
Anyone else up for a review?

davebx:
@nsoranzo The one in `package_hisat_0_1_5` is pretty significantly broken. Another solution would be to create `package_hisat_0_1_6`, which has a working script to extract splice sites.

@bgruening I'll have a look at the latest samtools, if it works as intended, I'll update the dependency.

bgruening:
@davebx the idea with `package_hisat_0_1_6` was postponed?

nsoranzo:
@davebx @bgruening I don't know enough about HISAT, but it probably makes sense to update to 0.1.6 now instead of having to remember to update extract_splice_sites.py separately.

jj-umn:
@davebx I started on a whist wrapper last August and it included most of the advanced options as well.  Do you want that for reference?   Do we have a datatype for splice junctions?

davebx:
@jj-umn Sure, and I don't know of a specific datatype for splice junctions.

davebx:
@bgruening I discovered that the official precompiled downloads for HISAT 0.1.6 are packaged in .zip files, which don't preserve the executable bit. I've contacted the author about this, and expect to hear back shortly.

bgruening:
If this does not work let's put it into the depot (-:

davebx:
Upon further review, the zip file provided by ccb.jhu does have the executable flag set, but the tool dependency installation process doesn't preserve it, because zipped files' external attributes need to be explicitly interpreted and set on the extracted files. I should have a pull request to modify that behavior in not too long.

bgruening:
@davebx I guess we still need to repack this into a tarball for a few releases until your patch is in every Galaxy release.

nsoranzo:
Isn't issuing a few `<action type="chmod">` easier?

davebx:
@nsoranzo Easier, yes, but it's a workaround instead of addressing the actual cause.

davebx:
@bgruening Yeah, which I'll take care of right now, it's as trivial as `unzip && tar czf`

nsoranzo:
@davebx Also repacking the tarball is a workaround! Anyway, the real fix is the one you mentioned in https://github.com/galaxyproject/tools-devteam/pull/137#issuecomment-108894537 .

davebx:
Repacked tarball: https://github.com/galaxyproject/tools-iuc/pull/178

bgruening:
@nsoranzo can we merge this?

data_managers/data_manager_hisat_index_builder/data_manager/hisat_index_builder.xml
```diff
@@ -0,0 +1,67 @@
+<tool id="hisat_index_builder_data_manager" name="HISAT index" tool_type="manage_data" version="1.0.0">
+    <description>builder</description>
+    <requirements>
+        <requirement type="package" version="0.1.5">hisat</requirement>
```
nsoranzo: s/0.1.5/0.1.6/

data_managers/data_manager_hisat_index_builder/tool_dependencies.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0"?>
+<tool_dependency>
+  <package name="hisat" version="0.1.5">
+      <repository name="package_hisat_0_1_5" owner="iuc" prior_installation_required="False" />
```
nsoranzo: `s/0.1.5/0.1.6/` and `s/package_hisat_0_1_5/package_hisat_0_1_6/`

This file is not properly indented.

tools/hisat/hisat.xml
```diff
@@ -0,0 +1,354 @@
...
+        #if str($input_format.paired.paired_selector) == 'paired':
+            -1 "${reads_f}" -2 "${reads_r}"
+        #else if str($input_format.paired.paired_selector) == 'paired_collection':
+            -1 "${input_format.paired.reads.forward}" -2 "${input_format.paired.reads.reverse}"
+        #else if str($input_format.paired.paired_selector) == 'paired_list':
+            #set forward_reads = []
+            #set reverse_reads = []
+            #for read in $input_format.paired.reads:
+                $forward_reads.append($read.forward)
+                $reverse_reads.append($read.reverse)
```
nsoranzo: Do these 2 lines need to be prefixed with `#silent`?

davebx: That doesn't appear to be necessary in this case.

tools/hisat/tool-data/hisat_indices.loc.sample
```diff
@@ -0,0 +1,37 @@
+# bowtie2_indices.loc.sample
+# This is a *.loc.sample file distributed with Galaxy that enables tools
+# to use a directory of indexed data files. This one is for Bowtie2 and Tophat2.
+# See the wiki: http://wiki.galaxyproject.org/Admin/NGS%20Local%20Setup
+# First create these data files and save them in your own data directory structure.
+# Then, create a bowtie_indices.loc file to use those indexes with tools.
+# Copy this file, save it with the same name (minus the .sample),
```
nsoranzo: This should reference HISAT instead of Bowtie2/Tophat2.

nsoranzo:
Thanks @davebx !

PR #138, opened 2015-05-21
Update cuffdiff to produce tabular output by default and optionally produce sqlite

davebx:


tool_collections/cufflinks/cuffdiff/cuffdiff_wrapper.xml
```diff
@@ -186,6 +187,74 @@
         </conditional>
     </inputs>
     <outputs>
+        <!-- Optional read group datasets. -->
+        <data format="tabular" name="isoforms_read_group" label="${tool.name} on ${on_string}: isoforms read group tracking" from_work_dir="isoforms.read_group_tracking" >
```
bgruening: @davebx @jmchilton could this go into a data collection?

bgruening:
@davebx why not offer a tool that extracts certain types of files/informations from the sqlite table?

tool_collections/cufflinks/cuffdiff/cuffdiff_wrapper.xml
jmchilton: One could do that but I would recommend strongly against - data collections should be homogeneous. These tabular files each have a different purpose - they are not just different samples for instance.

davebx:
@bgruening We do, under tools/cummerbund_to_tabular/

bgruening:
@davebx can I reuse cummerbund_to_tabular to get all different files this PR offers? If so why do we need this PR? :)

jxtx:
1) Since 2010 the cuffdiff wrapper has created the tabular outputs, which are what cuffdiff itself produces. Changing this not only breaks backward compatibility, but breaks the expected behavior. This is one of the most used tools on Galaxy main, and extensive training materials have been developed for it by many groups. When we can, we should try to maintain expected behavior as much as possible.

2) This tool is no longer cuffdiff, but cuffdiff + cummeRbund. The output of the tool is a specialized form that can *only* be used by one other tool in Galaxy, cummeRbund. To do any analysis other than visualization another tool is needed to restore the original outputs, which can be used by many other tools in Galaxy.

Both of these points seem fundamentally against Galaxy philosophy to me. I think the PR needs to go further and make these outputs the default as they always have been.

If we want to improve interfaces as tools change that is great, but we should do this incrementally, and we should ensure that old param_dicts can be migrated forward, even if this requires implementing new migration capabilities for new tool versions.

If we are concerned about having too many datasets in the history, we need to implement more grouping capabilities (perhaps extending collections) rather than just packing things into a proprietary datatype.

bgruening:
>1) Since 2010 the cuffdiff wrapper has created the tabular outputs, which are what cuffdiff itself produces. Changing this not only breaks backward compatibility, but breaks the expected behavior. This is one of the most used tools on Galaxy main, and extensive training materials have been developed for it by many groups. When we can, we should try to maintain expected behavior as much as possible.

Fair point. We have updated our documentation but I know this is major work.

> 2) This tool is no longer cuffdiff, but cuffdiff + cummeRbund. The output of the tool is a specialized form that can *only* be used by one other tool in Galaxy, cummeRbund. To do any analysis other than visualization another tool is needed to restore the original outputs, which can be used by many other tools in Galaxy.

I see your point here, but I don't think this is in general true. Aggregating informations from heavy compute jobs into a single dataset is imho a user-friendly way. With an additional tool you can decide which information your really need. Most of the cufflinks output is rarely used I would argue.

> Both of these points seem fundamentally against Galaxy philosophy to me. I think the PR needs to go further and make these outputs the default as they always have been.

Improving the usability is pro-Galaxy. I can only judge from my experiences during workshops - finding the correct cufflinks file was always error prone.

> If we want to improve interfaces as tools change that is great, but we should do this incrementally, and we should ensure that old param_dicts can be migrated forward, even if this requires implementing new migration capabilities for new tool versions.

Incrementally ... maybe don't enable tabular output by default? :)

> If we are concerned about having too many datasets in the history, we need to implement more grouping capabilities (perhaps extending collections) rather than just packing things into a proprietary datatype. 

Yes, please.

jxtx:
> I see your point here, but I don't think this is in general true. Aggregating informations from heavy compute jobs into a single dataset is imho a user-friendly way. With an additional tool you can decide which information your really need. Most of the cufflinks output is rarely used I would argue.

Even if that is true, I'd rather see this solved by implementing a general way to group datasets and pull out sub-datasets that would work with any tool. I'm *always* opposed to pragmatic hacks in tools when we could fix the problem at the framework level because tools are with us *forever*. 

> Incrementally ... maybe don't enable tabular output by default? :)

Perhaps, but the logical first increment would be start producing the cummeRbund file as well. Then we can start adding that to training, but existing materials will still work. This gives people a path to update their materials, rather than needing to change everything overnight.

bgruening:
>> I see your point here, but I don't think this is in general true. Aggregating informations from heavy compute jobs into a single dataset is imho a user-friendly way. With an additional tool you can decide which information your really need. Most of the cufflinks output is rarely used I would argue.

> Even if that is true, I'd rather see this solved by implementing a general way to group datasets and pull out sub-datasets that would work with any tool. I'm *always* opposed to pragmatic hacks in tools when we could fix the problem at the framework level because tools are with us *forever*. 

Sure, I'm all for it. `dataset collections` are awesome and something like this would be fantastic. See my initial question to John, where I ask if we could put this into a collection.

>> Incrementally ... maybe don't enable tabular output by default? :)

> Perhaps, but the logical first increment would be start producing the cummeRbund file as well. Then we can start adding that to training, but existing materials will still work. This gives people a path to update their materials, rather than needing to change everything overnight. 

I need to update my documentation now as well :) These wrappers are available since 2 month, so not sure what is the best way to handle such updates. 
Feel free to merge this as it is.

bgruening:
xref: https://biostar.usegalaxy.org/p/12455/

martenson:
Thank you Dave

PR #140, opened 2015-05-26
Update picard_FilterSamReads.xml

bgruening:
FilterSamReads can only take one input at a time, as far as I know.


martenson:
seems as a correct fix according to https://broadinstitute.github.io/picard/command-line-overview.html

PR #144, opened 2015-06-01
Adopt for cufflinks error message

bgruening:
Cufflinks can produce the following error message and this isn't catched by the current stdio tag.
```
Error running cufflinks.
return code = -11
Command line:
```

### DIFF
--- a/tool_collections/cufflinks/cuff_macros.xml
+++ b/tool_collections/cufflinks/cuff_macros.xml
@@ -10,8 +10,8 @@
     <stdio>
         <exit_code range="1:" />
         <exit_code range=":-1" />
-        <regex match="Error:" />
-        <regex match="Exception:" />
+        <regex match="Error" />
+        <regex match="Exception" />
     </stdio>
   </xml>
   <xml name="condition_inputs">


blankenberg:
I think the better fix is to correct cufflinks_wrapper.py to properly relay return codes, instead of silently eating them in e.g. stop_err(), or eliminate it completely. Looks like theres several issues with this wrapper script.

bgruening:
@blankenberg :+1:  on fixing the wrapper, or getting rid of it completely. For the time being the current situation is that a dataset turn green if this error occurs.

blankenberg:
Unfortunately, I don't have time for a proper fix to this issue and don't want to be a blocking factor, so I am OK with this as a fix for now. 

Green but really an error is a pretty big problem IMHO. Should it have a version bump?

bgruening:
@blankenberg thanks for letting us know. I don't think a version bump is needed. No previous "results" will change after this commit, if we define "results" as green datasets :)

bgruening:
@nsoranzo, version bump? :)

nsoranzo:
I'd say no, but may be debatable.

bgruening:
@blankenberg feel free to merge.

PR #149, opened 2015-06-12
Fix failed functional test

golharam:
The functional test for this tool fails.  Looking at the XML wrapper, it looks like $comments is invalid....it doesn't exist anywhere else in the xml file.  I know that comments can be specified multiple times however there is no way to get the comments fields to appear multiple times in the UI as far as I can tell.  This change allows the comment to be added and still pass.

### DIFF
--- a/tools/picard/picard_MarkDuplicatesWithMateCigar.xml
+++ b/tools/picard/picard_MarkDuplicatesWithMateCigar.xml
@@ -14,10 +14,8 @@
     OUTPUT="${outFile}"
     
     METRICS_FILE="${metrics_file}"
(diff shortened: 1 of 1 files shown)


nsoranzo:
:+1:

This is the same approach of picard_FastqToSam.xml .
The alternative solution may be to use a `<repeat>` as in picard_AddCommentsToBam.xml , picard_MarkDuplicates.xml and picard_MergeSamFiles.xml .

bgruening:
:+1:

bgruening:
Thanks @golharam!

PR #151, opened 2015-06-12
Correct line 43 of picard_MergeBamAlignment.xml

MatthewRalston:
Line 128 specifies a parameter named read2_aligned_bam from the read2_aligned_bams repeat block. On line 43, this parameter is referenced as "read1_aligned_bam." Whether or not this makes a difference is beyond my understanding of the internals, but consistency is desirable. Alternatively, if this is not managed behind the scenes by the galaxy library then this would have produced incorrect output in the past.


jmchilton:
Seems like a good fix to me. :+1:

bgruening:
I'm wondering if we should increase the version number?

nekrut:
@MatthewRalston: Pls, increase version and I'll merge.

martenson:
thanks @MatthewRalston

MatthewRalston:
Thank you all.

PR #152, opened 2015-06-12
Kraken update2

nekrut:
kraken actually works now. added kraken-mpa-report. still need to add tests

### DIFF
--- a/tool_collections/kraken/kraken_report/kraken-mpa-report.xml
+++ b/tool_collections/kraken/kraken_report/kraken-mpa-report.xml
@@ -1,0 +1,68 @@
+<tool id="kraken-mpa-report" name="Kraken-mpa-report" version="1.1.0">
+    <description>
+        view report of classification for multiple samples
+    </description>
+    <macros>
+        <import>macros.xml</import>
+    </macros>
+    <command>
+      <![CDATA[
+
+#for $input_count, $input_classification in enumerate( $classifications ):
+ ln -s "${input_classification.classification}" "sample_${input_count}" &&
+#end for
+
+kraken-mpa-report @INPUT_DATABASE@
+
+#for $input_count, $input_classification in enumerate( $classifications ):
+ "sample_${input_count}"
+#end for
+
+${show_zeros}
+${header_line}
+
+> "$output_report"
+
+
+	]]>
+    </command>
+    <inputs>
+      <repeat name="classifications" title="Classifications" min="1">
+	<param name="classification" type="data" format="tabular" label="Kraken output"/>
+      </repeat>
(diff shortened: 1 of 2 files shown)


davebx:
:+1:

tool_collections/kraken/kraken_report/kraken-mpa-report.xml
```diff
@@ -0,0 +1,68 @@
...
+${show_zeros}
+${header_line}
+
+> "$output_report"
+
+
+	]]>
+    </command>
+    <inputs>
+      <repeat name="classifications" title="Classifications" min="1">
```
bgruening: Can we use multiple="True" here?

tool_collections/kraken/kraken_report/kraken-mpa-report.xml
```diff
@@ -0,0 +1,68 @@
...
+ 4. A rank code, indicating (U)nclassified, (D)omain, (K)ingdom, (P)hylum, (C)lass, (O)rder, (F)amily, (G)enus, or (S)pecies. All other ranks are simply filled with a dash.
+ 5. NCBI taxonomy ID
+ 6. Indented scientific name
+
+The scientific names are indented using spaces, according to the tree
+structure specified by the taxonomy.
+    ]]>
+    </help>
+    <expand macro="version_command" />
+    <expand macro="requirements" />
```
bgruening: I think `planemo` will complain here. Requirements and stdio and version_command should be at the top.

tool_collections/kraken/kraken_report/kraken-mpa-report.xml
nekrut: we certainly should

tool_collections/kraken/kraken_report/kraken-mpa-report.xml
nekrut: I'll fix that

tool_collections/kraken/kraken_report/kraken-mpa-report.xml
nekrut: actually, I appears that we cannot as it triggers multiple jobs instead

bgruening: @nekrut why is this so? It shouldn’t ... iteration over the `multiple=True` should be equivalent with repeat, afaik.

nekrut: yes, this is how I did it in the latest commit

bgruening:
I'm going to merge this, but this tool really needs some tests.

PR #155, opened 2015-06-17
Cleanup Python files of gops tools.

nsoranzo:
All tests pass for me.


tool_collections/gops/basecoverage/operation_filter.py
```diff
@@ -1,8 +1,5 @@
 # runs after the job (and after the default post-filter)
-import os
-from galaxy import eggs
-from galaxy import jobs
-from galaxy.tools.parameters import DataToolParameter
+from galaxy.tools.parameters.basic import DataToolParameter
```
jmchilton: Is this imported from basic anywhere else in tools? I feel like ``galaxy.tools.parameters`` is a better public API entry point for this. basic is an okay name - but it is not as ... obviously correct as ``galaxy.tools.parameters``. Also basic has too much junk in it - we can probably never refactor it but in a perfect world I would like to so I would prefer not to have more things depend on it.

nsoranzo: No, it is imported only in gops tools. Anyway, I can revert this change if you prefer.

jmchilton: Especially if it is only imported here - I think this should be reverted. That means maybe someday we can refactor and breakup basic.py :).

nsoranzo: @jmchilton Done.

jmchilton:
:+1:

jmchilton:
Awesome - thanks for the fix and clean ups @nsoranzo.

Issue #160, opened 2015-06-18
No change when using option "unsorted" as sort order with picard_sortsam


jennaj:
The purpose of this option is unclear. Could we clarify it in the tool form?

Galaxy Tool Version 1.126.0

bgruening:
Hi Jen,
judging from the code it should `unsorted` is used and should work: https://github.com/galaxyproject/tools-devteam/blob/master/tools/picard/picard_SortSam.xml

The question is what is picard doing with this option :)

nsoranzo:
I think that this simply does not sort the input file and set the `SO` header of the output file to `unsorted`. Not very useful...

From the SAM specification: "SO: Sorting order of alignments. Valid values: unknown (default), unsorted, queryname and coordinate."

jennaj:
Hum, perhaps there is a use case for changing the header .. but I can't think of one! :)

bgruening:
@jennaj, @nsoranzo not sure if we should remove this option. But if this is to confusing for users you can review https://github.com/galaxyproject/tools-devteam/pull/165

jennaj:
I will let you all decide, just an observation. Maybe clarification in the help about what this option actually does is the best call?

bgruening:
Template for @jennaj 

* [ ] in the main tool shed
* [ ] installed successfully and tested via the UI with sample data on:
* [ ]  Test (https://test.galaxyproject.org) and Main (http://usegalaxy.org)
* [ ]  Local (default getgalaxy.org, master branch)
* [ ]  Cloudman (default cloudlaunch, latest master)

PR #165, opened 2015-06-18
remove "unsorted" option from picard SortSam

bgruening:
for @jennaj and https://github.com/galaxyproject/tools-devteam/issues/160

### DIFF
--- a/tools/picard/picard_SortSam.xml
+++ b/tools/picard/picard_SortSam.xml
@@ -1,4 +1,4 @@
-<tool id="picard_SortSam" name="SortSam" version="@TOOL_VERSION@.0">
+<tool id="picard_SortSam" name="SortSam" version="@TOOL_VERSION@.1">
   <description>sort SAM/BAM dataset</description>
   <macros>
     <import>picard_macros.xml</import>
@@ -21,7 +21,6 @@
     <param name="sort_order" type="select" display="radio" label="Sort order" help="SORT_ORDER; default=coordinate">
       <option value="coordinate" selected="True">Coordinate</option>
       <option value="queryname">Queryname</option>
-      <option value="unsorted">Unsorted</option>
     </param>
     <expand macro="VS" />
   </inputs>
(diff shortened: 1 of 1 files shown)


jennaj:
Optionally include help comments on the tool form that explain what this option does (modifies the SO: header). Either OK with me now that it has been clarified.

nsoranzo:
So we close this?

bgruening:
I'm fine with removing this option.

nsoranzo:
@nekrut opinions?

bgruening:
Can we merge this one as well?

bgruening:
@nekrut any ideas here?

bgruening:
ping @jennaj @nekrut

jennaj:
@bgruening my vote was always "yes" to remove it since it seems way simpler, but I was ok with the other (documenting what it did). but @nekrut and the devs have the final say. \

Sorry for late reply - I missed this somehow!

bgruening:
@jennaj @nekrut @nsoranzo I rebased the changes and I'm ready to merge this.

nsoranzo:
:+1:

jennaj:
As an informal way of tracking final testing, could we please leave this open and track when the updated tool moves through these steps: 

1) in the main tool shed
2) installed successfully and tested on:
  a) http://usegalaxy.org
  b) Local (default getgalaxy.org, master branch) 
  c) Cloudman (default cloudlaunch, running latest master)

After all completed, give the final "go" and close out the ticket.

Sound OK?

martenson:
@jennaj this is a Pull Request. Unless we actually close it nothing will change. Only after the close the changes will get merged into the codebase. We cannot 'leave it open'.

bgruening:
@jennaj: @martenson is right, this is not possible here. But it is possible in your original issue #160 
You have a fair point that we are missing to distribute these fixes/enhancement. @erasche is working on an automatic TTS upload. But installing on cloudman and Test is something only the devteam can do.

Have a look at #160. I have added a template how you can structure your issue with a checkbox to keep track of this process.

jennaj:
@bgruening excellent, I like the idea of tracking the resolution and final testing in the original ticket (which often describes the **observed** issue in detail w/ example data). The checkboxes are a simple way to track movement through these wrap-up steps. I may rearrange the contents a bit, all that was off the top of my head :) And suggestions from all are welcome.

Ideally, any user-side tool or important UI modifications would get some sort of default "Integration, Test, and Close" checklist. Install for main/test would be devteam of course, install local/cloud probably me along with UI testing on all three (but we can decide best path/who - could be different people depending on who knows what to look for in the testing).

We do all this anyway informally. Tracking to make sure all is done so users know when a tool is ready for use again was the primary goal.

PR #172, opened 2015-06-26
data_manager_sam_fasta_index_builder error on missing fasta file

blankenberg:
samtools faidx provides a 0 return code when the fasta input file is missing. This will cause the tool to properly enter an error state.


bgruening:
:+1:

PR #173, opened 2015-06-27
bwa - add dbKey to output bwa

kaktus42:
BWA tool now adds the dbKey of the reference sequence to the output bam.

PS: This repo is newer than what is on the toolshed, right? When do you update the toolshed?

### DIFF
--- a/tools/bwa/bwa-mem.xml
+++ b/tools/bwa/bwa-mem.xml
@@ -289,7 +289,9 @@
   </inputs>
 
   <outputs>
-    <data format="bam" name="bam_output" label="${tool.name} on ${on_string} (mapped reads in BAM format)"/>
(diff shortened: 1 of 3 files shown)


bgruening:
Nice! Could you add please a test for this function!

nsoranzo:
I think only `dbKeyActions` needs to be a macro, the other 2 are used only once.

kaktus42:
@bgruening And How would I do that?
@nsoranzo For me redundancy reduction is not the only reason to use macros. In this case e.g. for structuring

nsoranzo:
@kaktus42 I'm ok with that.

kaktus42:
ping @bgruening

bgruening:
Thanks @kaktus42!

PR #175, opened 2015-07-04
HISAT only accepts GTF, not GFF.

davebx:


bgruening:
Thanks Dave.

bgruening:
Btw. does hisat support "directional" as TopHat?

PR #176, opened 2015-07-06
update picard to version 1.136

kaktus42:
updates picard to version 1.135

when yould this be uploaded to toolshed?

### DIFF
--- a/packages/package_picard_1_135/.shed.yml
+++ b/packages/package_picard_1_135/.shed.yml
@@ -1,3 +1,8 @@
 owner: devteam
 name: package_picard_1_135_0
-    
+description: Contains a tool dependency definition that downloads and compiles version 1.135 of the Picard package.
+homepage_url: http://broadinstitute.github.io/picard/
+remote_repository_url: https://github.com/galaxyproject/tools-devteam/blob/master/packages/package_picard_1_135/
+type: tool_dependency_definition
+categories:
+- Tool Dependency Packages
--- a/packages/package_picard_1_135/.shed.yml
+++ b/packages/package_picard_1_135/.shed.yml
@@ -1,0 +1,3 @@
+owner: devteam
+name: package_picard_1_135_0
+    
--- a/packages/package_picard_1_135/.shed.yml
+++ b/packages/package_picard_1_135/.shed.yml
@@ -1,8 +1,0 @@
-owner: devteam
-name: package_picard_1_135_0
-description: Contains a tool dependency definition that downloads and compiles version 1.135 of the Picard package.
-homepage_url: http://broadinstitute.github.io/picard/
-remote_repository_url: https://github.com/galaxyproject/tools-devteam/blob/master/packages/package_picard_1_135/
-type: tool_dependency_definition
-categories:
-- Tool Dependency Packages
--- a/tools/picard/picard_AddCommentsToBam.xml
+++ b/tools/picard/picard_AddCommentsToBam.xml
@@ -1,4 +1,4 @@
-<tool name="AddCommentsToBam" id="picard_AddCommentsToBam" version="1.126.0">
+<tool name="AddCommentsToBam" id="picard_AddCommentsToBam" version="1.135">
   <description>add comments to BAM dataset</description>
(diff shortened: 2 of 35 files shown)


tools/picard/read_group_macros.xml
```diff
@@ -1 +0,0 @@
-../../macros/read_group_macros.xml
```
nsoranzo: This symlink should not be removed!

kaktus42:
@nsoranzo fixed accidently removed symlink and added shed.yml to package

tools/picard/picard_AddOrReplaceReadGroups.xml
```diff
@@ -1,4 +1,4 @@
-<tool name="AddOrReplaceReadGroups" id="picard_AddOrReplaceReadGroups" version="1.126.0-gx-1">
+<tool name="AddOrReplaceReadGroups" id="picard_AddOrReplaceReadGroups" version="1.135-gx-1">
```
martenson: I think we can safely abandon the ```-gx-1``` part. It is a relict of when we were not sure how to version.

martenson:
@kaktus42 looks very good! thank you very much for your contribution. I will test the installation and then merge this later.

bgruening:
@martenson can we merge this?

tools/picard/picard_AddCommentsToBam.xml
```diff
@@ -1,4 +1,4 @@
-<tool name="AddCommentsToBam" id="picard_AddCommentsToBam" version="1.126.0">
+<tool name="AddCommentsToBam" id="picard_AddCommentsToBam" version="1.135">
```
erasche: picard will update eventually, and as galaxy is using the underlying tool version, I don't suppose you'd be willing to encapsulate this in a macro?

E.g. https://github.com/galaxyproject/tools-iuc/blob/master/tools/bedtools/macros.xml#L8 is a macro which is then used like this: https://github.com/galaxyproject/tools-iuc/blob/master/tools/bedtools/bed12ToBed6.xml#L1

packages/package_picard_1_135/.shed.yml
```diff
@@ -0,0 +1,3 @@
+owner: devteam
+name: package_picard_1_135_0
+
```
erasche: Feel free to refuse this request too, but it'd be nice if this could be filled out like the IUC does. It'd help us automatically upload it. E.g. https://github.com/galaxyproject/tools-iuc/blob/master/packages/package_bedtools_2_19/.shed.yml and then copy the metadata from the existing picard package in the TS.

tools/picard/picard_AddCommentsToBam.xml
kaktus42: That makes it impossible to update single tools withing this tool “package”. Can't find where I've seen that before, but I would say, this should be discussed first somewhere else.

packages/package_picard_1_135/.shed.yml
kaktus42: Done with 9037c87

martenson:
:+1:

kaktus42:
In the meanwhile, there was a new version of picard published.
Were on 1.136 with 63cfaaf

erasche:
Excellent, though  63cfaaf is arguably an argument for https://github.com/galaxyproject/tools-devteam/pull/176#discussion_r34820101

:+1:

tools/picard/picard_AddCommentsToBam.xml
erasche: See the bed tool, it uses `version="@WRAPPER_VERSION@.0"`, each tool includes an individual postfix'd number which can be updated individually of the whole set which should depend consistently on a single version of picard.

erasche:
LGTM, merging.

erasche:
Will push packages/updates to MTS presently.

erasche:
In case anyone with devteam MTS credentials is reading this who isn't on IRC, please push these. I goofed and pushed under IUC, because I lack devteam TS creds.

PR #177, opened 2015-07-07
Add more environment variables to libxml2

JanKanis:
I had trouble using libxml2 from my blast2html tool which is written in python. Currently I'm using a private fork of package_libxml2_2_9_1 but I would prefer to use the mainline one. 

The changes are setting LD_LIBRARY_PATH, LIBRARY_PATH, C_INCLUDE_PATH and CPLUS_INCLUDE_PATH. That allows pip to find the right location of the binary and headers to install the lxml module in the venv that my tool builds.


bgruening:
@JanKanis wonderful! Thanks a bunch!

JanKanis:
Wow, fast response time.

PR #178, opened 2015-07-07
Make HISAT handle paired lists better.

davebx:


bgruening:
LGTK, @davebx can we use @jmchilton macros to handle collection more uniformly in a next version?

jmchilton:
:+1: - thanks for the fix!

jmchilton:
Not sure about the macros in the abstract - if you see specific tools with overlap though - we can do that.

PR #179, opened 2015-07-10
bwa - fixing malformed header because of readgroups

kaktus42:
the `\t` gets replaced with a real tab and messes up the run/bam file because the `@PG` line gets crappy.
Using `\\t`, like also recommended by bwa [1] fixes this.

[1] http://bio-bwa.sourceforge.net/bwa.shtml

### DIFF
--- a/tools/bwa/bwa-mem.xml
+++ b/tools/bwa/bwa-mem.xml
@@ -1,5 +1,5 @@
 <?xml version="1.0"?>
-<tool id="bwa_mem" name="Map with BWA-MEM" version="0.3">
+<tool id="bwa_mem" name="Map with BWA-MEM" version="0.4">
   <description>- map medium and long reads (&gt; 100 bp) against reference genome</description>
   <macros>
     <import>read_group_macros.xml</import>
--- a/tools/bwa/bwa.xml
+++ b/tools/bwa/bwa.xml
@@ -1,5 +1,5 @@
 <?xml version="1.0"?>
-<tool id="bwa" name="Map with BWA" version="0.3.0">
+<tool id="bwa" name="Map with BWA" version="0.4.0">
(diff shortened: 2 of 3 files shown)


kaktus42:
push

bgruening:
@kaktus42 where do you see this in the documentation? I only can find this here `’@RG\tID:foo\tSM:bar’`

kaktus42:
@bgruening that's what I'm talking about.
This is a bugfix, because the previous version was converted to this command:
`-R '@RG	ID:foo	SM:bar'`
and the command comes into `@PG`:
`@PG	ID:bwa [...]	-R @RG	ID:foo	SM:bar`
which gets confused because there is two times ID
with using `\\t`,
the command becomes `-R '@RG\tID:foo\tSM:bar'` like it should be.

bgruening:
@jennaj @jmchilton I do not use read groups so often, can we merge this. I trust @kaktus42 but I haven't tested it.

jennaj:
I vote for merging, putting on Test first, integration testing (using most current dev branch), then updating usegalaxy.org.

Does BWA-MEM also need to be updated?

kaktus42:
This changes all BWA tools that set read groups.
I tested it successfully if that matters.

jmchilton:
:+1: for the change - thanks for fix this issue @kaktus42. Can you update the version of any affected wrapper tool XML file? We have been updating the version of bug fixes like this and it looks like the read group changes have landed in the tool shed.

bgruening:
@kaktus42 awesome thanks and sorry for taking such a long time!

kaktus42:
:+1: when will you update toolshed?

bgruening:
@kaktus42 I can not update it. But usually it will be updated quite fast by the devteam. @davebx can you update this please?

PR #185, opened 2015-07-21
Fix for sam_to_bam when sam does not include header information

scholtalbers:
The previous cmd tried to run samtools sort first, which would fail on a sam without header. This fixes that.


bgruening:
@scholtalbers can you please increase the version number of the tool.

bgruening:
@scholtalbers can you please check your last commit? You reverted you initial changes?

scholtalbers:
yeah, I forgot that I reverted the changes to actually test that the added test fails with the original tool wrapper.

bgruening:
This looks good now. Anyone else? @erasche @nsoranzo @peterjc ?

bgruening:
Thanks @scholtalbers!

PR #190, opened 2015-07-28
Add extra environment vars to libxslt

JanKanis:
This is needed so that python can find the library. LD_LIBRARY_PATH is needed for runtime loading of the lxml module, the others to build the
lxml module correctly.

Without this lxml will build against the system libraries, so everything would appear to work if the libxslt/libxml2 development files are
installed.


JanKanis:
By the way, you recently merged a similar PR for libxml2. When will those changes find their way into the Galaxy tool shed?

bgruening:
Awesome! Thanks.
@davebx can you please update libxml and libxslt? Thanks :)

PR #196, opened 2015-07-30
Have package_cairo_1_12_14 use fontconfig as a dependency.

blankenberg:


bgruening:
Do we need this as LD_LIBRARY_PATH during tool runtime?

bgruening:
@blankenberg I'm ok with merging this. Can you please rebase it?

bgruening:
Oh damn, I spend a while to track this down again and now I find this again ... :)
@blankenberg I have created a new package here: https://github.com/galaxyproject/tools-iuc/pull/281/files#diff-dd359d50e1040724fe05887db7ebbb88

Actually, I have modified the fontconfig package to export more ENV so cairo will build with it automatically.

bgruening:
@blankenberg can you update this one to the recent changes in other packages?

bgruening:
ping @blankenberg we have updates in IUC for all these packages which should be more stable now. Maybe it worth to have a look at the IUC packages again and rebase this PR.
Thanks!

gregvonkuster:
@blankenberg @bgruening I need this package and it is not compiling for me on OS X.  I'll see if I can get it working, but don't wait for me if you are going to try to merge this for Linux.

gregvonkuster:
I've been running into so many problems trying to get this to compile on OS X though.  I'm beginning to think it may not be worth the effort.  The problem is this:
STDERR
configure: error: pkg-config >=  required but not found (http://pkgconfig.freedesktop.org/)
Googling says that OS X comes with a very old version of pkg-config, but I haven't had any luck getting any of them to compile - I've tried many of these versions:
http://pkgconfig.freedesktop.org/releases/
Is anyone using this package on OS X?

bgruening:
Sorry @gregvonkuster I can not help you here. I also think that maintaining OS-X is a burden for very small gain, decide for yourself if your tool should run on OS-X or not.

mvdbeek:
@bgruening @blankenberg 
Unless there is a problem with the current package I guess this PR should be closed, since fontconfig is now a dependency of cairo.
@gregvonkuster 
It works for me, but I did install pkg-config from homebrew.
This is a list of things I got in my current environment.
```
cloog
gcc
gdbm
gettext
git
glib
gmp
isl
libffi
libmpc
mpfr
node
openssl
p7zip
pkg-config
python
readline
sqlite
```
A `brew install pkg-config` should do it.
See also http://brew.sh/ for getting brew.
If that doesn't work it might also be a problem with the received environmental variables,
which can result in an unrelated pkg-config error.

gregvonkuster:
@mvdbeek Thanks, I was kinda thinking of going that route (brew install pkg-config), but wasn't sure if we could make it a dependency.  I'll probably go with this to make progress though.

mvdbeek:
@gregvonkuster I'd love to see this as a dependency, since pkg-config is also not part of the build-essentials package on ubuntu, and we had some problems in the past (https://github.com/galaxyproject/tools-iuc/issues/407)

PR #198, opened 2015-07-31
Krona

nekrut:
Krona tool for visualization of taxonomy datasets in galaxy. Next I will work on converting kraken data into richer galaxy taxonomy format.


tools/krona/krona.xml
```diff
@@ -0,0 +1,87 @@
...
+            <param name="combine_inputs" value="False"/>
+            <output name="output" file="krona_test1.html" ftype="html" lines_diff="30"/>
+        </test>
+    </tests>
+    <help>
+        <![CDATA[
+
+**What it does**
+
+Creates an ineractive Krona_ chart based on Galaxy taxonomic representations.
```
bgruening: Why `Galaxy` here?

bgruening:
LGTM!
@nekrut join the IUC with your tools!

tools/krona/tool_dependencies.xml
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0"?>
+<tool_dependency>
+	<package name="perl" version="5.18.1">
+        <repository name="package_perl_5_18" owner="iuc" prior_installation_required="True" />
+    </package>
+
```
bgruening: How does this relates to https://github.com/galaxyproject/tools-iuc/pull/232?

nekrut: directly. it depends on it

tools/krona/krona.xml
nekrut: Not sure I understand. Krona has a tool `ktImportGalaxy` while turns galaxy taxonomy datatype into krona html document.

tools/krona/tool_dependencies.xml
bgruening: But you haven’t defined a dependency here, you are installing it by your own.
You also have not defined a requirement in your tool, afaik.

nekrut: Perhaps a part of the confusion is the decision to put deps into tools-uic and actual tools into tools-devteam. Seems a bit too complicated for me though

tools/krona/tool_dependencies.xml
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0"?>
+<tool_dependency>
+	<package name="perl" version="5.18.1">
```
bgruening: Indentation is wrong here.

tools/krona/tool_dependencies.xml
bgruening: :)
You need to include something like:
```
<package name="krona" version="2.5">
        <repository name="package_krona_2_5" owner="iuc" />
</package>
```
The perl requirement can go away, imho.
To stop the confusion commit your tools to IUC as well :)

tools/krona/krona.xml
bgruening: Not sure this detail is interesting for a user. I would argue that a user don't care about the underlying representation details.
Leave it like it is, it's fine with me.

tools/krona/tool_dependencies.xml
davebx: I'll fix that in post, too.

nsoranzo:
This is mostly overlapping with https://toolshed.g2.bx.psu.edu/view/crs4/taxonomy_krona_chart which is available since 2014-04-11. I think updating the old one would better serve the 75 users that have already installed it.

nekrut:
@nsoranzo - you guys are too productive. let me compare and see

nekrut:
Closed. Instead see https://github.com/galaxyproject/tools-iuc/pull/245

PR #204, opened 2015-08-04
Transfer R's dependencies to the IUC.

davebx:


erasche:
It's "dead code", why not just remove it? They'll always be available in git history.

martenson:
@erasche I might be old fashioned but there is available and 'available' in my book. It does not cost us anything to have them in /legacy and it improves accessibility.

erasche:
@martenson I see it more as duplication than accessibility. The TS will say "look here for the source" which has been updated to the tools-iuc repo, and if anyone is really interested in going back from that they'll easily see these two referenced PRs. Does anyone really want to look through the git history for archival versions of packages? Is it worth the clutter?

martenson:
@erasche I see your point but I am willing to maintain the /legacy folder in spite of it being a clutter.

nsoranzo:
I'm with @erasche on this.

jmchilton:
:+1: for move to ``/legacy`` - there shouldn't be stuff in the tool shed that is not located in source code somewhere even if we don't plan to update it. git history is not easy to find or reference.

bgruening:
I also like to have it around, just because it's easy to grep in and get new ideas. For me this is the best documentation and I'm teaching this (shame on me).

erasche:
I really believe y'all are trying to support a minority use case. 

> there shouldn't be stuff in the tool shed that is not located in source code somewhere

I mean... it isn't going away. It's there permanently. Would y'all accept a PR version of 204/240 which retains edit history? I will gladly do that if we don't have to have a `legacy/` folder.

> git history is not easy to find or reference

1. Browse TS, find cairo package.
2. Click on Repository link
3. "Hmm, I wonder how this package used to be built?" (who does this? really?)
4. Instead of going through TS history which links real code to actual package versions, choose to go through git history. 
5. Click blame/history links in github view of that package.
5. Github shows you it was part of https://github.com/galaxyproject/tools-iuc/pull/240 which xrefs to #204 
6. See that they were deleted. Browse repo at parent commit
7. Git history/blame from there.

> grep in and get new ideas

As for grepping, unless you're doing some magic, you're only grepping the latest, best versions (which is good). If you're grepping, just include tools-iuc to your tools-devteam search.

martenson:
@erasche We already have a legacy folder (https://github.com/galaxyproject/tools-devteam/tree/master/legacy) and there are tools that were replaced by other tools (bamleftalign). Not all legacy tools cases are as simple as move from devteam -> iuc

I think trying to retain full edit history is too much work when compared to simple /legacy move (and e.g. in the case of bamleftalign also impossible).

erasche:
@martenson okay, since this is community consensus, I'll merge.

martenson:
Thank you Eric.

PR #207, opened 2015-08-05
Use github magic to download tarball.

davebx:
Per @bgruening's suggestion.


bgruening:
As always thanks Dave, this looks to me much cleaner! Reminds we to go over my old packages and use this as well ;)

PR #208, opened 2015-08-06
Add data manager to download NCBI taxonomy

davebx:


davebx:
XREF: https://github.com/nekrut/tools-devteam/pull/3

bgruening:
General question. Do we need to have the name field here, or can we simply default to a readable date format?
From a usability point of view the only thing I care is the fetch-date? Right?

@blankenberg @davebx what do you think?
I would always enter the date into the name field :)

davebx:
@bgruening That's essentially how I was picturing the name field being used, as a place for a galaxy admin to enter the download date.

data_managers/data_manager_fetch_ncbi_taxonomy/data_manager/ncbi_taxonomy_fetcher.xml
```diff
@@ -0,0 +1,19 @@
+<?xml version="1.0"?>
+<tool id="ncbi_taxonomy_fetcher" name="NCBI" tool_type="manage_data" version="1.0.0">
+    <description>taxonomy downloader</description>
+    <stdio>
+        <exit_code description="Error" level="fatal" range="1:" />
+    </stdio>
+    <command>
+    <![CDATA[
+        curl -o taxdump.tar.gz ${taxonomy_url} && tar xzf taxdump.tar.gz && python $__tool_directory__/data_manager.py --name "${database_name}" --out "${out_file}"
+    ]]>
```
bgruening: `curl` is not installed on every system, so I think we need to depend on https://github.com/galaxyproject/tools-iuc/blob/master/packages/package_libcurl_7_35/tool_dependencies.xml

davebx: Does that install the curl executable as well?

bgruening:
@davebx can we simply default to str(datetime.date.today()) if the user has not entered anything, with a nice help text under this parameter?

data_managers/data_manager_fetch_ncbi_taxonomy/data_manager/ncbi_taxonomy_fetcher.xml
bgruening: Very good question, let see if I have it installed on Galaxy-FR.

bgruening: Ha, have it and yes in /bin there is a curl binary.

bgruening: We should remove the ssl dependency on this package.

davebx: Awesome, I'll update the dependencies right away.

davebx:
@bgruening Good idea.

bgruening:
:+1: Thanks @davebx and sorry for the delay.

PR #212, opened 2015-08-26
bamtools_filter remove old static var

martenson:
and move the files to reflect how they are in the TS


bgruening:
@martenson I somehow liked `static` but maybe I was just used to it. Should we add a sentence to the best-practice guide and mention `/images`?

martenson:
@bgruening the joke is that it does not matter where you put it it will always get somehow mapped to root folder; or at least that is what my testing shows

bgruening:
Hopefully this is a feature, but it makes me a little bit nervous. What happen if you have two images with the same name. One in test-data one somewhere else?
Nevertheless, we should agree on a standard where to put help-multimedia.
1: `./images`
2: `./img/`
3: `./media/`
4: `./static/media`

martenson:
@bgruening I tested it right now - when named the same it will use the one in root folder

bgruening:
Ok! Thanks! reverse engineering :)
Any preferred location to store media?

martenson:
It seems the method `get_absolute_path_to_file_in_repository` is searching for the path in the repo dir and causes this behavior, we might want to enhance it to handle multiple files properly (if it is a path resolve it first before searching)

davebx:
:+1: For this workaround, until someone has the time and energy to fix the root issue.

PR #221, opened 2015-09-11
remove size attribute from tools

bgruening:


erasche:
@bgruening normally I'm happy to see tools cleaned up, but I noticed the second bullet point today when I was commenting on nitesh's PR:

https://github.com/galaxyproject/tools-devteam/blob/master/CONTRIBUTING.md#what-not-to-contribute

nsoranzo:
I think the second bullet point should be removed, this repository contains a lot of actively used tools (bowtie2, cufflinks, samtools just to name a few).

guerler:
:+1:

erasche:
@nsoranzo, sure. But the rule seems to be there so we aren't generating revisions of old tools which aren't meaningful upgrades? Whether or not they're actively used is separate to "stylistic revisions"

And yes, :+

martenson:
What is the reasoning behind removing these params? New tool form does not use it?

erasche:
@martenson yes, https://github.com/galaxyproject/tools-iuc/pull/268

bgruening:
@martenson this attribute is not used anymore and we should clean things up.

I'm with @nsoranzo here, this bullet point is sub-optimal. I understand why this is there and I think it's pointless to get external contributions to clean up things, so it's good to communicate this.
My motivation for this PR was that this repository is still used a lot by others to learn tool wrapping and coping stuff. So I think it's good to remove cruft in all our example repositories. 
I hope that no new installable revision is created for these changes.

jmchilton:
Reluctant :+1:  I guess. Lets leave the second bullet point alone though, we spend too much time working on old tools no one is ever going to use again.

https://twitter.com/jxtx/status/644189567613038592

jxtx:
Is this the right solution? Should the new tool form be updated to use this instead? I think it is important to remember that the tool config language is meant to be generic, not linked to any particular UI implementation. So, if this is a valuable piece of information for describing the parameter we should keep it whether the UI uses it or not.

(I don't have an opinion on whether it is useful though, so +/- 0)

martenson:
I am with @jxtx here regarding UI neutrality.
I can imagine that if I was to design a from-the-scratch toolform and had these size params available I would make use of them.

bgruening:
I see this repository and the IUC one as a great archive and documentation. The size attribute is copied over to new tools and produces unexpected behaviour (no impact). At least this happened in one of your workshops.
From my side I'm glad that the size attribute is gone. It was not nice to work with. I always had the feeling that the size I have chosen was wrong :)

erasche:
The votes are in favour of merging, and it's been sitting open far too long. There's obviously work to be done in the future to make the tools more UI-neutral, but given how non-uniformly the `size` attribute was used I'm in favour of merging this as-is. Thanks for the cleanup @bgruening

Issue #228, opened 2015-09-24
Cuffdiff "Condition names" with spaces result in an unclear error


jennaj:
Problem: Condition names that are entered with spaces were once a problem, then resolved (escaped), but are now problematic again. @blankenberg suggested that the wrapper just needs a bit of tuning.

Trello with detail: https://trello.com/c/MiKnwCZQ

Error for reference:
```

job info:
This job failed for reasons that could not be determined.
Please click the bug icon to report this problem if you need help.

```

jennaj:
Ok now plus is on main. Thanks for addressing this, helps many in community! 

Sorry for delay in confirming, lost ticket :-/ Have a new procedure now for tracking known issues to integration test.

PR #229, opened 2015-09-24
Update cuff_macros.xml

bgruening:
fixes https://github.com/galaxyproject/tools-devteam/issues/228

@blankenberg what do you think about this fix?


blankenberg:
another alternative: https://github.com/galaxyproject/tools-devteam/pull/231

bgruening:
@blankenberg please decide what we want to do here. I don't think pipe is a good solution, but feel free to decide this.

bgruening:
ping @blankenberg

nsoranzo:
This has been solved by commit https://github.com/galaxyproject/tools-devteam/commit/8434d197771a311f78b50e3c5293e135df28f978 .

PR #233, opened 2015-09-29
Update download urls for cufflinks source code

peterjc:
These are 404:

- http://cufflinks.cbcb.umd.edu/downloads/cufflinks-2.1.1.tar.gz
- http://cufflinks.cbcb.umd.edu/downloads/cufflinks-2.2.1.tar.gz

Only the links for the binary packages had previously been updated, see https://github.com/galaxyproject/tools-devteam/commit/f00c3318751dab6a608688a8706798a27cd883ba (@martenson) and https://github.com/galaxyproject/tools-devteam/commit/804f9e6a17882f1353fe0a82692ca887c5a6e3da (@bgruening)

Found while testing ``planemo dependencies_script`` - see https://github.com/galaxyproject/planemo/pull/310


martenson:
good catch, thanks @peterjc

PR #235, opened 2015-09-29
Update tophat source code URLs

peterjc:
These are 404:

- http://tophat.cbcb.umd.edu/downloads/tophat-1.4.0.tar.gz
- http://tophat.cbcb.umd.edu/downloads/tophat-2.0.9.tar.gz

Note there is a ``tophat-1.4.0.1.tar.gz`` available on http://ccb.jhu.edu/software/tophat/downloads/ which I suspect was a minor Linux fix, but this is not mentioned here: https://ccb.jhu.edu/software/tophat/index.shtml

Found while testing planemo dependencies_script - see galaxyproject/planemo#310


martenson:
was just looking into this because of cufflinks links :)

PR #236, opened 2015-09-30
Makes no-contain/no-overlap option labels consistent with text

bwlang:
I think the label was backwards before this fix (it read like you have to check the box to allow mate overlaps)

From the bowtie manual:

"Bowtie 2's default behavior is to consider overlapping and containing as being consistent with concordant alignment. By default, dovetailing is considered inconsistent with concordant alignment."


nsoranzo:
:+1:

bgruening:
Thanks @bwlang!!!

PR #238, opened 2015-10-05
Errror when not using "optional_inputs" flag with 'report_monomorphic flag'

Takadonet:
Fixing issue when 'optional_inputs' is not selected and then complain it cannot find 'report_monomorphic'.  parameter. Moved it to be under all other parameter that only show up when 'optional_inputs' is selected.

Example of stacktrace

  File "/Applications/galaxy/production/galaxy-dist/lib/galaxy/jobs/runners/__init__.py", line 163, in prepare_job
    job_wrapper.prepare()
  File "/Applications/galaxy/production/galaxy-dist/lib/galaxy/jobs/__init__.py", line 859, in prepare
    self.command_line, self.extra_filenames, self.environment_variables = tool_evaluator.build()
  File "/Applications/galaxy/production/galaxy-dist/lib/galaxy/tools/evaluation.py", line 422, in build
    raise e
NotFound: cannot find 'report_monomorphic' while searching for 'options_type.optional_inputs.report_monomorphic'

### DIFF
--- a/tools/freebayes/freebayes.xml
+++ b/tools/freebayes/freebayes.xml
@@ -354,7 +354,7 @@
         <param name="reference_allele_selector" type="boolean" truevalue="set" falsevalue="do_not_set" label="Use reference allele?" help="Sets --use-reference-allele and --reference-quality options  " />
         <when value="set">
           <param name="Z" type="boolean" truevalue="-Z" falsevalue="" checked="False" label="Include the reference allele in the analysis as if it is another sample from the same population" help="-Z --use-reference-allele; default=False" />
-          <param name="reference_quality" type="text" size="8" value="100,60" label="Assign mapping quality of MQ (100) to the reference allele at each site and base quality of BQ (60)" help="--reference-quality; default=100,60  " />
+          <param name="reference_quality" type="text" value="100,60" label="Assign mapping quality of MQ (100) to the reference allele at each site and base quality of BQ (60)" help="--reference-quality; default=100,60  " />
         </when>
         <when value="do_not_set">
            <!-- do nothing -->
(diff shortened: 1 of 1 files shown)


tools/freebayes/freebayes.xml
```diff
@@ -351,7 +354,7 @@
         <param name="reference_allele_selector" type="boolean" truevalue="set" falsevalue="do_not_set" label="Use reference allele?" help="Sets --use-reference-allele and --reference-quality options  " />
         <when value="set">
           <param name="Z" type="boolean" truevalue="-Z" falsevalue="" checked="False" label="Include the reference allele in the analysis as if it is another sample from the same population" help="-Z --use-reference-allele; default=False" />
-          <param name="reference_quality" type="text" value="100,60" label="Assign mapping quality of MQ (100) to the reference allele at each site and base quality of BQ (60)" help="--reference-quality; default=100,60  " />
+          <param name="reference_quality" type="text" size="8" value="100,60" label="Assign mapping quality of MQ (100) to the reference allele at each site and base quality of BQ (60)" help="--reference-quality; default=100,60  " />
```
blankenberg: We've removed and stop using the size attribute for text inputs.

blankenberg:
Probably want a version bump with this.

tools/freebayes/freebayes.xml
```diff
@@ -80,7 +80,11 @@
     #elif str( $options_type.options_type_selector ) == "full":
  
         #if $options_type.optional_inputs.optional_inputs_selector:
-        
+       
+	  #if $options_type.optional_inputs.report_monomorphic
+	    ${options_type.optional_inputs.report_monomorphic}
+	  #end if
```
nsoranzo: You don't need the `#if ...` and `#end if`, just the middle line as it was before.

Takadonet:
Update version number , removed size attribute for text field and removed #if and #end if statement for report_monomorphic.

nsoranzo:
:+1:

bgruening:
LGTM as well.

PR #239, opened 2015-10-06
add repository dependency to fetch genome data manager

bgruening:
@blankenberg do you think this PR is useful? 
I find myself often installing top level data managers (bowtie, bwa) without the underlying ones. A repository dependency could help here.


blankenberg:
I'm really not a fan of having the dependency at this level. I do totally get why it would be helpful though. 

Can we do something similar by creating several layers of hierarchical suites? (making suites stand out better during e.g. tool search/install would be nice as well)

bgruening:
This will also work :) Do you have any suggestions how to name and structure these new suites?

bgruening:
@blankenberg how do we want to proceed here? If you create one example I can follow :)

martenson:
I want to migrate all DMs out of this repo, can we close this? @bgruening @blankenberg

PR #240, opened 2015-10-06
Fix zcat on OS X.

nsoranzo:
`zcat -f file.dat` fails on OS X with something like `file not found file.dat.Z` . Thanks @anilthanki for reporting.

Also:
- quote `$input` and `$output` in `<command>`
- use `format_source` attribute instead of `format` for `<data>`
- fix some test cases
- whitespace fixes.

I've successfully tested all tools with planemo except for `snpfreq` (not tested).


bgruening:
@nsoranzo looks good to me. I'm wondering if we should depend on gnu-coreutils to prevent such errors.

nsoranzo:
@bgruening Maybe, but I think that for these old tools it's simpler to just fix the error in this way.

bgruening:
Ok, anyone wants to comment here. Does anyone think we need to increase the version number?

bgruening:
@martenson can we push this tool, please?

Issue #244, opened 2015-10-10
picard_CollectAlignmentSummaryMetrics.xml invalid value for MetricAccumulationLevel


scholtalbers:
When selecting all, the tool erros with this message (picard 1.126):
ERROR: 'ALL_READS,SAMPLE,LIBRARY,READ_GROUP' is not a valid value for MetricAccumulationLevel.

From the help:
METRIC_ACCUMULATION_LEVEL=MetricAccumulationLevel
LEVEL=MetricAccumulationLevel 
The level(s) at which to accumulate metrics.    Possible values: {ALL_READS, SAMPLE, LIBRARY, READ_GROUP} **This option may be specified 0 or more times**. This option can be set to 'null' to clear the default list. 

So I assume the list should be split to create "METRIC_ACCUMULATION_LEVEL=ALLREADS METRIC_ACCUMULATION_LEVEL=SAMPLE" etc.

bgruening:
@scholtalbers can you test this PR?

PR #245, opened 2015-10-10
Change METRIC_ACCUMULATION_LEVEL to a list of multiple values

bgruening:
fixes: https://github.com/galaxyproject/tools-devteam/issues/244

### DIFF
--- a/tools/picard/picard_CollectAlignmentSummaryMetrics.xml
+++ b/tools/picard/picard_CollectAlignmentSummaryMetrics.xml
@@ -4,18 +4,21 @@
     <import>picard_macros.xml</import>
   </macros>
   <expand macro="requirements" />
+  <stdio>
+    <exit_code range="1:"  level="fatal"/>
+  </stdio>
   <command>
     @java_options@
     ##set up input files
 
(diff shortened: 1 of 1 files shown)


tools/picard/picard_CollectAlignmentSummaryMetrics.xml
```diff
@@ -24,7 +24,9 @@
     #for $sequence in $adapters:
         ADAPTER_SEQUENCE="${sequence.adapter}"
     #end for
-    METRIC_ACCUMULATION_LEVEL="${metric_accumulation_level}"
+    #for $level in ${metric_accumulation_level}:
```
scholtalbers: Hi @bgruening, this does not work for me (galaxy@15.07). 
*Long-form placeholders - ${}, $(), $[], etc. are not valid inside expressions.* ==> should be ``in $metric_accumulation_level``. 
But changing that results in:
``TypeError: 'SelectToolParameterWrapper' object is not iterable``

bgruening:
Tests are passing for me now.

scholtalbers:
Yes, this works, thx!

martenson:
updated on MTS

PR #247, opened 2015-10-14
Update Cut tool to match distribution.

blankenberg:
see also: https://github.com/galaxyproject/galaxy/pull/913

### DIFF
--- a/tools/cut_columns/cutWrapper.pl
+++ b/tools/cut_columns/cutWrapper.pl
@@ -9,6 +9,9 @@ my @in = ();
 my @out = ();
 my $command = "";
 my $field = 0;
+my $start = 0;
+my $end = 0;
+my $i = 0;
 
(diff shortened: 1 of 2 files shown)


bgruening:
:+1: 
Why do we have this wrapper in two places?

bgruening:
ping @blankenberg

blankenberg:
@bgruening The text utils were originally migrated to the toolshed and removed from the distribution, but then added back to the distribution. We should keep both copies in sync, or remove one. I have another set of changes that adds header options and step-size to ranges, that I'd like to push out as well.

nsoranzo:
@blankenberg We can move the present version here to the `legacy/` subdirectory.

bgruening:
I will merge this. @blankenberg please tell us if we should deprecate this tool here.

Issue #251, opened 2015-10-19
Remove sort_by from all reference select boxes


scholtalbers:
I would like to keep control of how my local data, particular the genome references, appear in all tools and the  ``<filter type="sort_by" column="2" />`` gets in the way. Could it just be removed from all wrappers?

Btw. I'm assuming here that the available reference list is shown in the same order as they are under the admin local_data interface, which reflect the order in the ``tool-data/*.loc`` files

PR #253, opened 2015-10-22
Remove sort_by for build/indexes

scholtalbers:
If you guys think this is a good idea this removes sort_by from all tools (except for legacy/bwa).
Relates: #251

### DIFF
--- a/tools/bowtie2/bowtie2_wrapper.xml
+++ b/tools/bowtie2/bowtie2_wrapper.xml
@@ -268,7 +268,6 @@
           <when value="indexed">
             <param name="index" type="select" label="Select reference genome" help="If your genome of interest is not listed, contact the Galaxy team">
               <options from_data_table="bowtie2_indexes">
-                <filter type="sort_by" column="2"/>
                 <validator type="no_options" message="No indexes are available for the selected input dataset"/>
               </options>
             </param>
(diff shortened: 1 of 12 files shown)


bgruening:
@scholtalbers any reason to remove this?
I'm +/- 0 on this. It is kind of a documentation and tell others how to sort for other columns.

scholtalbers:
@bgruening I have some legacy build keys (different nomenclature) that I cannot or want to remove due to stored workflows and histories. I prepended their labels/names with (legacy) and put them at the bottom of the .loc files. For some tools this works, for others that have sort_by this doesn't and they appear at the top. I could prepend them differently, or maybe there is a filter I can place?(i did not look into this option yet) 
But I guess the main point is that the behavior can be different between tools and as an admin I would like to control which builds are shown first - those that are mainly used.

bgruening:
ping @blankenberg

blankenberg:
Now that we can type to search within these select elements, the alphabetical sorting is not nearly as important as before -- where the ordering in the flat files usually just reflected the 'random' ordering that they were created and added.

However, while scrolling it is nice to e.g. see all of the 'Human' builds together.

I'm like +.00001 on this, since an admin could manually reorder the original files to get the same outcome as the sort option (although this would be a manual step needed to get back to 'old look') and the searching feature in the select lists easing finding selections in large lists.

scholtalbers:
@blankenberg, could there be a default sorting when fetching data from .loc
files - which would also be reflected on the admin "local data" pages? This
would at least make it uniform (unless.. someone has a different sort_by in
the tool xml of course).

bgruening:
@scholtalbers any string feeling about this one or can we close it?

scholtalbers:
I don't mind if you close it. Although I do still feel that the sorting of the references/builds should not be the responsibility of the tool :).

bgruening:
@scholtalbers thats probably true. But than a github issue on the Galaxy main repo is the right place :)
Thanks!

PR #254, opened 2015-10-22
Add option to Bowtie & Bowtie2 to keep statistics from stderr

pjbriggs:
By default the `bowtie_wrapper` and `bowtie2` tools don't keep the stderr output from the underlying `bowtie` and `bowtie2` programs, however this contains useful statistics on the mapping.

This PR adds an option to both tools that, when selected, saves these statistics as a history item when the tools are run. (There are also new test cases, however I haven't been able to get these or the existing tests to pass.)

Adding this functionality has previously been discussed on this Trello card:

https://trello.com/c/TdYxdbkm

which also contains my original patch for the`bowtie_wrapper`. I've returned to this now I have users who are interested in having this functionality in our local Galaxy instance, however the thread on Trello suggests that others might also find it useful.

Thanks for considering these changes, all feedback welcome.

### DIFF
--- a/tools/bowtie2/bowtie2_wrapper.xml
+++ b/tools/bowtie2/bowtie2_wrapper.xml
@@ -435,7 +435,7 @@
                 <param name="sam_opt" type="boolean" truevalue="true" falsevalue="false" label="Would you like the output to be a SAM file" help="By default, the output from this Bowtie2 wrapper is a sorted BAM file."/>
             </when>
         </conditional>
-        <param name="save_mapping_stats" type="boolean" truevalue="True" falsevalue="False" checked="False" label="Save the bowtie2 mapping statistics to the history" />
+        <param name="save_mapping_stats" type="boolean" checked="False" label="Save the bowtie2 mapping statistics to the history" />
     </inputs>
 
     <!-- define outputs -->
--- a/tools/bowtie2/bowtie2_wrapper.xml
+++ b/tools/bowtie2/bowtie2_wrapper.xml
@@ -167,6 +167,11 @@
             ${analysis_type.cline}
         #end if
 
+        ## mapping stats (i.e. stderr from bowtie2)
+        #if $save_mapping_stats
+	  2&gt; "$mapping_stats"
+        #end if
+
         ## output file
         #if ( str( $analysis_type.analysis_type_selector ) != "full" or str( $analysis_type.sam_opt ) != "true" ):
           | samtools view -Su - | samtools sort -o - - &gt; $output
@@ -430,6 +435,7 @@
                 <param name="sam_opt" type="boolean" truevalue="true" falsevalue="false" label="Would you like the output to be a SAM file" help="By default, the output from this Bowtie2 wrapper is a sorted BAM file."/>
             </when>
         </conditional>
+        <param name="save_mapping_stats" type="boolean" truevalue="True" falsevalue="False" checked="False" label="Save the bowtie2 mapping statistics to the history" />
     </inputs>
 
     <!-- define outputs -->
@@ -517,6 +523,9 @@
             </conditional>
           </actions>
         </data>
+        <data format="txt" name="mapping_stats" label="${tool.name} on ${on_string}: mapping stats">
+          <filter>save_mapping_stats is True</filter>
+        </data>
 
     </outputs>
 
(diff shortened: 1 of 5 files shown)


davebx:
I like this, but I feel the tool wrapper version should be incremented when the tool's behavior changes.

pjbriggs:
@davebx thanks for the feedback - I've bumped the version numbers on both.

bgruening:
@pjbriggs what are the problems with the tests?
This looks good to me, thanks!

pjbriggs:
@bgruening thanks - re the tests, I couldn't get them to pass using planemo, as the comparison of the output `bam`'s against the reference files failed. I suspect that the content of the fis okay and the problem is to do with either the `bam` files from bowtie being unsorted and/or with the `bam` comparison implemented in the Galaxy test code - I'm not 100% sure which.

(I did look at the comparison code in `test/base/twilltestcase.py` to see if I could get this to work but ran out of time before I could make much progress. I guess this is a separate issue from the bowtie wrapper updates?)

nekrut:
It is possible that the bam file is different in your tests because bowtie prints something run-specific (e.g., path).

mvdbeek:
> @bgruening thanks - re the tests, I couldn't get them to pass using planemo, as the comparison of the output bam's against the reference files failed. I suspect that the content of the files are okay and the problem is to do with either the bam files from bowtie being unsorted and/or with the bam comparison implemented in the Galaxy test code - I'm not 100% sure which.

It is the conversion from bam to sam which is not always producing sorted output (See https://github.com/galaxyproject/galaxy/issues/745). I'd say let's merge it.
Ping @bgruening @davebx

bgruening:
Oh this makes sense. Unfortunately, we need @pjbriggs to rebase it again. Sorry Peter!

pjbriggs:
@nekrut apologies - I had been meaning to get back to this for a while, thanks for your comments.

@mvdbeek @bgruening thanks, I've had a go at rebasing the branch. The tool version in this branch needs bumping again (the conflict was with the version also being updated in master, to match the `bowtie` package version) - it's currently `2.2.6`, I guess it should go to `2.2.6.1`?

Let me know your preference and I'll make the appropriate change in my branch.

mvdbeek:
@pjbriggs yes, `2.2.6.1` would be best.

pjbriggs:
@mvdbeek done - let me know if there's anything more to do on this.

tools/bowtie2/bowtie2_wrapper.xml
```diff
@@ -430,6 +435,7 @@
                 <param name="sam_opt" type="boolean" truevalue="true" falsevalue="false" label="Would you like the output to be a SAM file" help="By default, the output from this Bowtie2 wrapper is a sorted BAM file."/>
             </when>
         </conditional>
+        <param name="save_mapping_stats" type="boolean" truevalue="True" falsevalue="False" checked="False" label="Save the bowtie2 mapping statistics to the history" />
```
nsoranzo: You don't need to specify `truevalue` and `falsevalue`.

nsoranzo:
It seems you didn't rebase correctly, there are changes to other files not related to Bowtie.

pjbriggs:
@nsoranzo hrm yes you're right. I will try and revert the incorrect attempt (hopefully it won't make things worse...)

tools/bowtie2/bowtie2_wrapper.xml
pjbriggs: @nsoranzo thanks - this was consistent with the definition of the existing `boolean` parameters in the wrappers, but I've updated my changes to take them out.

pjbriggs:
@nsoranzo have reverted and reattempted the rebase, hopefully looking okay this time.

tools/bowtie_wrappers/bowtie_wrapper.py
```diff
@@ -432,6 +434,11 @@ def __main__():
             tmp_stderr.close()
             if returncode != 0:
                 raise Exception, stderr
+            elif options.output_mapping_stats is not None:
+                # Write stderr (containing the mapping statistics) to a named file
+                mapping_stats = open( options.output_mapping_stats, 'w')
+                mapping_stats.write( stderr )
+                mapping_stats.close()
```
nsoranzo: Here you can use:
```
                with open(options.output_mapping_stats, 'w') as mapping_stats:
                    mapping_stats.write( stderr )
```

pjbriggs: @nsoranzo thanks should be addressed now.

nsoranzo:
:+1: 

Thanks @pjbriggs!

pjbriggs:
@nsoranzo @mvdbeek @bgruening thanks for all your feedback - is there anything left for me to do now on this or can the changes be merged? Pls let me know.

bgruening:
@pjbriggs thanks a lot!

pjbriggs:
@bgruening thanks!

Issue #3, opened 2014-09-29
Add citations to all the tools.


jmchilton:
- [ ] add_value > tools/add_value
- [ ] analyze_covariates > tool_collections/gatk/analyze_covariates
- [ ] annotation_profiler > tools/annotation_profiler
- [x] bamleftalign > tools/freebayes
- [x] bam_to_sam > tool_collections/samtools/bam_to_sam
- [ ] basecoverage > tool_collections/gops/basecoverage
- [ ] best_regression_subsets > tools/best_regression_subsets
- [ ] blat_coverage_report > tools/blat_coverage_report
- [ ] blat_mapping > tools/blat_mapping
- [x] bowtie2 > tools/bowtie2
- [x] bowtie_color_wrappers > tools/bowtie_color_wrappers
- [x] bowtie_wrappers > tools/bowtie_wrappers
- [x] bwa_wrappers > tools/bwa_wrappers
- [ ] canonical_correlation_analysis > tools/canonical_correlation_analysis
- [ ] categorize_elements_satisfying_criteria > tools/categorize_elements_satisfying_criteria
- [ ] ccat > tools/ccat
- [ ] change_case > tools/change_case
- [x] clustalw > tools/clustalw
- [ ] cluster > tool_collections/gops/cluster
- [ ] column_maker > tools/column_maker
- [ ] complement > tool_collections/gops/complement
- [ ] compute_motif_frequencies_for_all_motifs > tools/compute_motif_frequencies_for_all_motifs
- [ ] compute_motifs_frequency > tools/compute_motifs_frequency
- [ ] compute_q_values > tools/compute_q_values
- [ ] concat > tool_collections/gops/concat
- [ ] condense_characters > tools/condense_characters
- [ ] convert_characters > tools/convert_characters
- [ ] convert_solid_color2nuc > tools/convert_solid_color2nuc
- [ ] correlation > tools/correlation
- [ ] count_covariates > tool_collections/gatk/count_covariates
- [ ] count_gff_features > tools/count_gff_features
- [ ] coverage > tool_collections/gops/coverage
- [ ] ctd_batch > tools/ctd_batch
- [x] cuffcompare > tool_collections/cufflinks/cuffcompare
- [x] cuffdiff > tool_collections/cufflinks/cuffdiff
- [x] cufflinks > tool_collections/cufflinks/cufflinks
- [x] cuffmerge > tool_collections/cufflinks/cuffmerge
- [ ] cut_columns > tools/cut_columns
[...]

PR #258, opened 2015-10-24
Support keep_first_word option

erasche:
Wildly useful tool, however keeping the whole description isn't great. The substring option is nice however only works for fasta files with identical length IDs which is ... basically 0% of the ones we encounter.

This exposes the existing option in code to just use the first "word", (so the fasta ID but not the fasta description).


tools/fasta_compute_length/fasta_compute_length.xml
```diff
@@ -1,9 +1,13 @@
 <tool id="fasta_compute_length" name="Compute sequence length" version="1.0.0">
 	<description></description>
-	<command interpreter="python">fasta_compute_length.py $input $output $keep_first</command>
+	<command interpreter="python">fasta_compute_length.py $input $output $keep_first $keep_first_word</command>
 	<inputs>
 		<param name="input" type="data" format="fasta" label="Compute length for these sequences"/>
 		<param name="keep_first" type="integer" value="0" label="How many title characters to keep?" help="'0' = keep the whole thing"/>
+        <param name="keep_first_word" type="boolean" truevalue="id_only" falsevalue="id_and_desc"
```
nsoranzo: It seems this file is mixing tabs with spaces, so here the indentation looks wrong.

tools/fasta_compute_length/fasta_compute_length.xml
```diff
@@ -1,9 +1,13 @@
 <tool id="fasta_compute_length" name="Compute sequence length" version="1.0.0">
```
nsoranzo: Increase the version number.

nsoranzo:
Where are the test data for this tool? :open_mouth:

erasche:
@nsoranzo I get the impression it uses the built in Galaxy test data? Dunno. :/

bgruening:
@erasche fixed @nsoranzo concers.

test-data is here: https://github.com/galaxyproject/galaxy-test-data
And we all probably have a cached version on our hardrive :)
The question is do we want to include it here, or is it enough to trigger planemo to download these test-data?

peterjc:
A long overdue enhancement :)

jmchilton:
Please don't add empty citations, hides a valid a problem :). Lets add:

```
<citations>
  <citation type="doi">10.1093/bioinformatics/btq281</citation>
</citations>
```

bgruening:
@jmchilton Just to clarify, if there is no citation we can/should add an empty citations tag, right?

jmchilton:
@bgruening Yes, but it should be fairly rare right? I'd also recommend a comment in these cases:

```<citations><!-- No citations because... --></citations>```

bgruening:
Not sure, more or less all tools with a own python script don't have a citation. I was not aware of this publication, so sorry for the trouble. Can the devteam go through all the tools and add citations?

jmchilton:
No trouble at all. I'd say most Python tools should cite something, the github repository they are in, an in preparation publication, etc....

As for citing all the tools, someday I guess :) Issue #3 (https://github.com/galaxyproject/tools-devteam/issues/3).

nsoranzo:
@bgruening @erasche All tools in this repository (except for this, `dwt_var_perfeature` and `velvet`) have the test-data included, and I think this is how it was always supposed to be.

I'm going to open a separate PR for the other 2 tools, but it would be nice to fix `fasta_compute_length` here.

nsoranzo:
:+1:

bgruening:
Great thanks!

PR #265, opened 2015-10-31
Freebayes TNG

bgruening:
This PR add a merge and split functionality to freebayes.
In my initial tests I was able to run freebayes on 90 Samples, on 24 Cores in 24h.

This is the important commit:
https://github.com/galaxyproject/tools-devteam/commit/db6051bdd5df909d0f8e0331dcbb178783e722f2

I tried splitting by chromosomes which makes the merging easier, but splitting by regions makes it 30% faster, so I think the more complicated merging is worth it.

@nekrut @blankenberg I would like to get your feedback on this.

Ideally we could use the `parallelization` mode of Galaxy. But since this in buggy and do not support the range functionality I was playing around with such a solution.


nekrut:
This is wonderful :+1:

tools/freebayes/freebayes.xml
```diff
@@ -1,529 +1,627 @@
...
     #else:
         #set $reference_fasta_filename = str( $reference_source.ref_file.fields.path )
     #end if
-    
+
     #for $bam_count, $input_bam in enumerate( $reference_source.input_bams ):
-        ln -s "${input_bam.input_bam}" "localbam_${bam_count}.bam" &amp;&amp;
-        ln -s "${input_bam.input_bam.metadata.bam_index}" "localbam_${bam_count}.bam.bai" &amp;&amp;
+        ln -s "${input_bam}" "b_${bam_count}.bam" &&
+        ln -s "${input_bam.metadata.bam_index}" "b_${bam_count}.bam.bai" &&
```
bgruening: I'm a little bit worried here about the maximum character limit of a shell. We can probably hit this with 1000 of files ...

nekrut: what is the shell limit, 2621440?

bgruening: >$ getconf ARG_MAX 
2097152

bgruening: btw. this is the reason I changes `localbam_` to `b_`. If we hit this limit we should talk to the freebayes developers I guess.

GuyReeves:
Hi Björn 
When will this make it to the toolshed? I would love to use it to save me having to slice bams. Thanks Guy

bgruening:
Still not sure about the merging step. I would pass the decision over the @blankenberg :) It works for us the quality values can change if they are in the overlapping regions.

GuyReeves:
I had wondered about how you might deal with overlaps.  For me I guess it will be fine.  Might it be worth thinking about having a user option where only whole chromosomes are sent out as jobs.  This might solve this problem with minimal downsides if your organism  has  more chromosomes that you have nodes. just a thought.

bgruening:
Yeah, at first I had an option to just split by chromosomes, but this was still 30% slower than this one.
Ideally this moves into a framework feature. @blankenberg will say me what to do :)

jmchilton:
Is it possible to expose the region argument alone - so someone could use a generic tool for producing a collection of bed files from a bam file and then just map freebayes over that and then merge the results? It may be too much work for the simplest case - but it would provide a lot of flexibility for advanced usages.

bgruening:
@jmchilton yes we can do a lot of these stuff, but I think this should be handled ideally on a framework level and not exposed to the user. It makes tools just too complicated and a normal user does not know how the cluster is structured and what kind of splitting procedure fits.

Until the framework is fixed I'm leaning to the split-by-chromosome approach. In this we do not have the VCF problem and it's still way way faster what we have currently.

jmchilton:
@bgruening Hmm... I'm not convinced it is a framework issue and not a tooling issue. If it is broken into jobs the Galaxy can funnel these into the correct place. There are aspects to that that one may want to tweak from cluster to cluster, but these are not as important or difficult to do as tweaking the tool inputs from application to application.

I'm still a fan of exposing the region file argument and parallelizing that way - I think it is the future.

bgruening:
@jmchilton how do you want to split it into multiple jobs from one tool execution without data collections. We want to have X jobs generated and distributed over the cluster not only X processes on one node. I don't see how this can work currently.
You could split a BAM file into a data collection and than run freebayes on the collection. This can be done today with the tools we have. I don't think this is what we should aim for if we can do the spitting on a framework level - just because it is so much easier for the user. 

Also please note the dependency on gnu-parallel to make this work in this PR.

bgruening:
ping @blankenberg how do we want to proceed here?
This PR is working but the quality values do slightly change because of the splitting by regions. If we split by chromosome this actually works and test-data stays the same.

Also we might want to update this to freebayes 1.x and use the splitting script they ship?

bgruening:
@blankenberg @nekrut @GuyReeves It's out of WIP. Can anyone review it.
It's conda only! But tested!

bgruening:
@jgoecks you also seemed to be interested in this.

blankenberg:
Looks ok to me. I do wonder how significant the difference in quality values is and what impact it can have. Can we at least add a note to the help about this? Generally you wouldn't expect to get a different output based upon a backend issue, such as galaxy slots being set to 1, 10, 100, etc.

bgruening:
@blankenberg what do you mean? I have disabled the splitting by regions, it should only split by chromosomes now? Isn't it?

blankenberg:
@bgruening I was going off of the previous comments and the modified test data. Although, the changes in values could be be due to differences in freebayes versions?

bgruening:
Yeah, this is my guess.

martenson:
@bgruening @blankenberg does this PR make https://github.com/galaxyproject/tools-devteam/pull/360 obsolete?

bgruening:
Not really. I'm waiting for a comment from @Takadonet and will insert the `-f` if needed.

lparsons:
Couple of questions:

1. Are there plans to release this to main toolshed?
2. I seem to be having a problem where population model settings are ignored. Is there an issue setting `truevalue="set"` in the input parameter and checking for `== "True"` in the command (cheetah)?

bgruening:
@lparsons I think this should be pushed to the MTS as all test pass. But it nice to get more testers and you probably found a bug. Can you fix it?
@martenson can you upload it if this is fixed?

martenson:
@bgruening sure, just ping me in this (or any other) thread whenever, I will push it

jgoecks:
Sorry to be so late on this, but can we simply use the freebayes-parallel script rather than roll our over solution? 

https://github.com/ekg/freebayes/blob/master/scripts/freebayes-parallel

I have this working on our local cluster and it works well: using 16 cores, it takes ~15 minutes to call variants across an exome. I can do a WIP pull request with my changes if it would be useful.

bgruening:
Please do ... This work started way earlier than this script.

bgruening:
Jeremy also please note that we also tried the region approach but had difficulties with changing quality values because of non-overlapping regions. Last time I checked this was also true for this script. For this reason we changed back to chromosome splitting, which should be save.

peterjc:
I wonder if and where my colleagues shared their parallel Freebayes wrapper, mentioned in https://bmcbioinformatics.biomedcentral.com/articles/10.1186/s12859-015-0801-z

> To speed up the SNP calling in FreeBayes, we produced a Java SE 7/SAMtools 0.1.18 [19] wrapper around it that splits and parallelises the job across multiple nodes and processors of a compute cluster. This allowed the jobs to run in a fraction of the time that would otherwise have been required. This is achieved by querying the list of contigs, discarding those that have no reads mapped to them, splitting the remainder into discrete regions that can be processed independently by FreeBayes, before finally concatenating the results back together into a single VCF (Variant Call Format) file.

I've asked: https://twitter.com/pjacock/status/771984349784834048

PR #267, opened 2015-11-02
add suites shedyamls

martenson:
includes fixes for @nsoranzo in the previous PR

Suites are generally troublesome because here they are not tied to the exact revisions of tools. I will go ahead and do some testing with this.

### DIFF
--- a/suites/suite_gatk_1_4/.shed.yml
+++ b/suites/suite_gatk_1_4/.shed.yml
@@ -1,4 +1,14 @@
-# repository published to https://toolshed.g2.bx.psu.edu/repos/devteam/suite_gatk_1_4
-owner: devteam
+categories:
+- Sequence Analysis
+description: A suite of Galaxy utilities associated with version 1.4 of the GATK package.
+long_description: |
+  The Genome Analysis Toolkit or GATK is a software package developed
+  at the Broad Institute to analyse next-generation resequencing data.  This repository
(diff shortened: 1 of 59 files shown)


martenson:
We might want to consider reverting https://github.com/galaxyproject/tools-devteam/commit/c2c801a6dd10a500b7e13d13f5f131e384bdad48

the way suites are versioned does not play well with the lack of versions in the actual repo dependencies.xml

bgruening:
@martenson the revision should be filled during upload? Isn't this the case?

martenson:
@bgruening it will fill the latest which breaks the versioning of suites

bgruening:
True ... but suites without a version number are supposed to install the latest versions?
Sorry, seems that I messed this up.

martenson:
I will revert part of #30 in a separate PR. Please review this one independently.

martenson:
created PR for it here: https://github.com/galaxyproject/tools-devteam/pull/269

martenson:
will merge today unless objections arise

suites/suite_gatk_1_4/.shed.yml
```diff
@@ -1,4 +1,14 @@
-# repository published to https://toolshed.g2.bx.psu.edu/repos/devteam/suite_gatk_1_4
-owner: devteam
+categories:
+- Sequence Analysis
```
nsoranzo: s/Sequence/Variant/

PR #269, opened 2015-11-03
revert stripping the changeset_revisions from versioned suites

martenson:
need to check whether Planemo automatically strips these on upload or not


martenson:
will merge today unless objections arise

PR #274, opened 2015-11-05
VarScan 2: added citation

yhoogstrate:
And some other tweaks to make planemo pass


nsoranzo:
I guess that would be a good time to also remove the citation from the ``<help>`` section and to update the 2 links with ``s|http://varscan.sourceforge.net/|http://dkoboldt.github.io/varscan/|``

PR #278, opened 2015-11-09
replace symlinked folders with symlinked files

martenson:
as planemo does not resolve symlinked folders (yet)


nsoranzo:
@martenson: Should we open a Planemo bug for this?

jmchilton:
@nsoranzo Enhancement :)

nsoranzo:
Changed to issue :)

PR #279, opened 2015-11-09
A tool for filtering reads by length and NM/length ratio based on a

nekrut:
…utility written by Nick Stoler. We will need this for IE paper.


nekrut:
Forgot deps and requirements. Do not look at yet

bgruening:
I will close this now as we have bam_filter on the TS.

PR #280, opened 2015-11-10
Clarify cd-hit-dup help that chimeric filter only available for single reads.

### LINKED ISSUE #273: cd-hit-dup "filter_chimeras" conditional
Why is this conditional i sonly limited to single (non-paired) reads?

```python

<when value="paired">
                <param name="fastq_input1" type="data" format="fastqsanger,fasta" label="Select first set of reads" help="Specify dataset with forward reads"/>
                <param name="fastq_input2" type="data" format="fastqsanger,fasta" label="Select second set of reads" help="Specify dataset with reverse reads"/>
            </when>
            <when value="single">
                <param name="fastq_input1" type="data" format="fastqsanger,fasta" label="Select reads" help="Specify dataset with single reads"/>
                <conditional name="filter_chimeras">
                    <param name="filter_chimeras_selector" type="select" label="Filter out chimeric clusters">
                        <option value="true">Yes</option>
                        <option value="false" selected="True">No</option>
                    </param>
                    <when value="true">
                        <param name="min_chimeric_length" type="integer" value="30" min="20" label="Minimum length of common sequence shared between a chimeric read and each of its parents" help="-s"/>
                        <param name="abundance_cutoff" type="integer" value="1" min="1" label="Abundance cutoff" help="-a; Tool Author recommend default of 2, but this would require the chimera itself to need 2 copies"/>
                        <param name="abundance_ratio" type="integer" value="1" min="1" label="Abundance ratio between a parent read and a chimeric read" help="-b"/>
                        <param name="dissimilarity_control" type="integer" value="1" min="1" label="Dissimilarity control for chimeric filtering" help="-p"/>
                    </when>
                    <when value="false">
                        <!-- do nothing here -->
                    </when>
                </conditional>
            </when>
```

blankenberg:
xref https://github.com/galaxyproject/tools-devteam/issues/273
ping @nekrut 
closes #273

### DIFF
--- a/tools/cd_hit_dup/cd_hit_dup.xml
+++ b/tools/cd_hit_dup/cd_hit_dup.xml
@@ -98,6 +98,10 @@
 
 cd-hit-dup is a simple tool for removing duplicates from sequencing reads, with optional step to detect and remove chimeric reads. 
 
+.. class:: infomark
+
+**TIP:** Chimeric filtering is only available on single-ended, or joined paired-end, reads.
+
 **Options**
 
 cd-hit-dup provides a number of options to tune how the duplicates are removed::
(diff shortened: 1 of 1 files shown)


bgruening:
I'm not sure why the test is failing, but this looks good to me.

PR #294, opened 2015-11-24
Allow FASTQ trimmer to accept negative values for right-hand offset.

blankenberg:
This will allow removing all sequence to the right of the absolute value of the provided negative position. 

IOW: a negative value will do the same as the standard python slice with the positive value.

ping @nekrut


martenson:
:+1:

bgruening:
:+1: waiting for @nekrut to merge.

PR #300, opened 2015-12-01
Remove double ampersand that remained after switching to autoconf

mvdbeek:
Same as https://github.com/galaxyproject/tools-iuc/pull/452
Ping @martenson


martenson:
updated in MTS, thanks @mvdbeek

PR #303, opened 2015-12-03
fix #302

### LINKED ISSUE #302: weblogo3 tool reports wrong number of sequences to stdout
One of our users reported the following today:

> Clicking on the output from that tool (on the right edge of the screen) reveals information about the search.
>
> Eg:  # weblogo input /mnt/silver/galaxy/containers/galaxy-qa/stable/database/files/013/dataset_13858.dat has 2845133 sequences all of length 10

> I noted, in my Excel manipulations that the reported number of sequences (as above) is always 1 smaller than the sum of the 4 nucleotides reported (for each sequence position) by the Sequence
Logo Tool.

It looks like an off-by-one error in [this line](https://github.com/galaxyproject/tools-devteam/blob/8bac50ca5572ea388181835e417b2db09fa27c1a/tools/weblogo3/rgWebLogo3.py#L108) of `rgWebLogo3.py`. `lasti` should be `lasti + 1` since the index is zero-based.

bgruening:
fixes: https://github.com/galaxyproject/tools-devteam/issues/302

### DIFF
--- a/tools/weblogo3/rgWebLogo3.py
+++ b/tools/weblogo3/rgWebLogo3.py
@@ -105,7 +105,7 @@ class WL3:
                 if seqlen <> flen:
                     print >> sys.stderr,self.badseq % self.opts.input
                     sys.exit(1)
-        return '# weblogo input %s has %d sequences all of length %d' % (self.opts.input,lasti,flen)
+        return '# weblogo input %s has %d sequences all of length %d' % (self.opts.input, lasti+1, flen)
 
 
     def run(self):


mvdbeek:
@bgruening so shouldn't this result in updated test-data ?!

mvdbeek:
nevermind, didn't see it only concerned to stdout, not the actual output file.
:+1:

mvdbeek:
@martenson can you upload this to the toolsheds?

bgruening:
Thanks for merging!

martenson:
@mvdbeek done! thanks a lot guys!

PR #310, opened 2015-12-09
add a list of all build-in functions

bgruening:
Not sure how to represent such a bulk of items ... a list seems to long to me.


blankenberg:
Looks good to me. If we were to do a list, we could use a table with explanations next to each item, but I think this is nice and compact.

PR #314, opened 2015-12-14
Fix package in tool_dependencies.xml for fastq_paired_end_joiner.

blankenberg:


martenson:
the package does not exist yet, but I see it in the IUC repo

PR #316, opened 2015-12-15
Fail the job if bwa index fails.

### LINKED ISSUE #312: Bwa-mem issues
- [x] stderr regex is less than ideal—a missing index gives a "fail" message but the job succeeds
- [ ] the output dataset build is not set when there is an entry in the loc file in the data manager directory but not in the loc file in the tool directory—this may be a framework issue.

nsoranzo:
Also use CDATA in `<command>`.

Partially fix #312.

Testing by an OS X user would be appreciated.

### DIFF
--- a/tools/bwa/bwa-mem.xml
+++ b/tools/bwa/bwa-mem.xml
@@ -1,5 +1,5 @@
 <?xml version="1.0"?>
-<tool id="bwa_mem" name="Map with BWA-MEM" version="0.4.2">
+<tool id="bwa_mem" name="Map with BWA-MEM" version="0.7.12">
   <description>- map medium and long reads (&gt; 100 bp) against reference genome</description>
   <macros>
     <import>read_group_macros.xml</import>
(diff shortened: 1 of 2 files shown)


nsoranzo:
Ping @jgoecks.

bgruening:
@mvdbeek since you are an OS-X user :)

mvdbeek:
hmm, so how do i make an index job fail?

mvdbeek:
before the update killing bwa in the indexing phase still resulted in a green dataset, with the update it properly turns red (OS X). :+1:

nsoranzo:
I've updated the pull request with an additional bug fix, i.e. to run the stat command on the reference dataset, not the symbolic link. Before the bug fix, the reference was always indexed with the IS algorithm.

@mvdbeek Can you check that this still works under OS X?

mvdbeek:
hmm, still working with small files (5 MB), with bigger files (50MB) as reference I get
```
  File "/Users/marius/src/galaxy/lib/galaxy/jobs/runners/local.py", line 128, in queue_job
    job_wrapper.finish( stdout, stderr, exit_code )
  File "/Users/marius/src/galaxy/lib/galaxy/jobs/__init__.py", line 1206, in finish
    dataset.datatype.set_meta( dataset, overwrite=False )
  File "/Users/marius/src/galaxy/lib/galaxy/datatypes/binary.py", line 314, in set_meta
    exit_code = subprocess.call( args=command, stderr=open( stderr_name, 'wb' ) )
  File "/usr/local/Cellar/python/2.7.10_2/Frameworks/Python.framework/Versions/2.7/lib/python2.7/subprocess.py", line 522, in call
    return Popen(*popenargs, **kwargs).wait()
  File "/usr/local/Cellar/python/2.7.10_2/Frameworks/Python.framework/Versions/2.7/lib/python2.7/subprocess.py", line 710, in __init__
    errread, errwrite)
  File "/usr/local/Cellar/python/2.7.10_2/Frameworks/Python.framework/Versions/2.7/lib/python2.7/subprocess.py", line 1335, in _execute_child
    raise child_exception
OSError: [Errno 2] No such file or directory
```
This is on the latest dev, so could be a dev problem as well. Downgrade to 15.10 is not so easy, since I run into problems with the mako egg ...

nsoranzo:
@mvdbeek This seems connected with setting metadata, so I don't think it's connected with this PR, but if you manage to test on a 15.10 (maybe a fresh Galaxy clone would be easier?), I would be more confident :)

mvdbeek:
Yep, I got 15.10 back to work with a clean venv, it continues to work fine. :+1:

mvdbeek:
Also the size comparison works fine
```
Reference genome size is 50999816 bytes, generating BWA index with bwtsw algorithm
```
(I lowered the 2000000000 for testing ...)

nsoranzo:
@mvdbeek Thanks for the fast and thorough testing (and for the merge)!

@martenson I think this is worth a new release on the MTS, if you have time.

martenson:
@nsoranzo done
thank you fellas! @mvdbeek

PR #320, opened 2016-01-15
BWA: Add param for reference indexing algorithm.

nsoranzo:
My users have been bitten again by the failure of BWA indexing when using a big reference genome with long sequences from the history.

```
[bwa_index] Pack FASTA... 26.13 sec
[bwa_index] Construct BWT for the packed sequence...
[is_bwt] Failed to allocate -7621388068 bytes at is.c line 211: Cannot allocate memory
```

With this patch the trick of calculating the genome size is substituted by a new param which let the user choose the indexing algorithm and by defaults let BWA decide which is the best algorithm for the genome, which seems to work better than the genome size criterion.


pvanheus:
Seems like a sensible extension of the bwa tool wrapper. May it be merged!

PR #322, opened 2016-01-18
freetype also need zlib

lecorguille:
Hi,

zlib rule them all.

So, I add a dependency to zlib.

Cheers


bgruening:
Thanks @lecorguille!

PR #325, opened 2016-01-21
Update samtools_mpileup to symlink bam and bai

jj-umn:
The input bam files need to have the associated indexes.  So using the model from the freebayes tool, symlink the inputs bams ans their associated indexes into the job working directory.


bgruening:
@jj-umn can you increase the version number please?

jj-umn:
An additional issue can occur when mpileup creates a bcf output when the Bcf datatype attempts to run bcftools index in set_meta when the bcf is uncompressed.

PR #327, opened 2016-02-01
Remove gffread optional

martenson:
change at L212 - the rest is whitespace

PR #328, opened 2016-02-01
Revert "Remove gffread optional"

martenson:
Reverts galaxyproject/tools-devteam#327


jmchilton:
Sorry, I closed this because this is a good change, I just don't think it is the right fix for the problem. I think we need to fix Galaxy.

martenson:
@jmchilton agreed

PR #332, opened 2016-02-05
tophat_fusion_post: additional parameters

jsh58:
Additional parameters added to wrapper.  Use of CDATA, best practices.


bgruening:
I like this!
Could you also run `planemo lint` on it.
ping  @blankenberg 

Thanks  @jsh58!

jsh58:
Updated tool wrapper -- reordered sections, added citation, added test.  Now passes planemo lint.  

However, the test does NOT work.  The wrapper uses the $__get_data_table_entry__() function and the 'dbkey' value of the input BAM file.  With the test dataset (test-data/accepted_hits.bam), the value of $accepted_hits.dbkey is not set, and I don't know how to set it.  With it unset, the blast_db_path and Bowtie indexes are not established, and the tool fails to run.

blankenberg:
```<param name="accepted_hits" value="accepted_hits.bam" dbkey="some_value" ftype="bam"/>```

jsh58:
That still fails the test, for me.

bgruening:
@blankenberg does the tests ever worked? Testing indices was not possible a long time. Merge anyway?

PR #334, opened 2016-02-10
changed display of params

matthdsm:
Changed display of params to radio button to increase the visibility of possible option, current version only shows an empty field with a "check/uncheck" box.

### DIFF
--- a/tools/table_annovar/table_annovar.xml
+++ b/tools/table_annovar/table_annovar.xml
@@ -63,21 +63,21 @@
         <validator type="unspecified_build" />
     </param>
 
-    <param name="gene_anns" type="select" display="radio" multiple="True" optional="True" label="Gene Annotations" help="" >
+    <param name="gene_anns" type="select" display="checkboxes" multiple="True" optional="True" label="Gene Annotations" help="" >
         <options from_data_table="annovar_indexes">
             <filter type="data_meta" key="dbkey" ref="input" column="1"/>
             <filter type="static_value" name="type" value="gene_ann" column="2"/>
         </options>
     </param>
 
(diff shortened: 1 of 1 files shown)


tools/table_annovar/table_annovar.xml
```diff
@@ -63,21 +63,21 @@
         <validator type="unspecified_build" />
     </param>
 
-    <param name="gene_anns" type="select" multiple="True" optional="True" label="Gene Annotations" help="" >
+    <param name="gene_anns" type="select" display="radio" multiple="True" optional="True" label="Gene Annotations" help="" >
```
nsoranzo: I think `display="checkboxes"` would be better. Radio buttons are normally used for mutually exclusive options, while here multiple selections are explicitly allowed: https://www.nngroup.com/articles/checkboxes-vs-radio-buttons/

matthdsm: I agree, would be more clear that way. Changed it!

Thanks for the input!
M

matthdsm:
On an unrelated note,
I'm trying to avoid to have to edit PATH of my galaxy user, but I can't seem to figure out where to put the env.sh file to set the environmental value.
Any help or resource? I'm a bit confused by the wiki entry at the [wiki](https://wiki.galaxyproject.org/Admin/Config/ToolDependencies)

Thanks
M

nsoranzo:
@matthdsm The following path may work: `YOUR_TOOL_DEP_DIR/environment_settings/SCRIPT_PATH/devteam/table_annovar/08b003ee9db7/env.sh` , but the revision (08b003ee9db7) will change at every tool repository update.

matthdsm:
Thanks! I forgot I could find those files there!
M

PR #336, opened 2016-02-11
changed usage of SCRIPT_PATH variable.

matthdsm:
Instead of pointing the SCRIPT_PATH to the installation dir, let it point to the tool-dependency dir where the user has installed the annovar scripts.
This variable can be set by the user in the env.sh file.

### DIFF
--- a/tools/table_annovar/table_annovar.xml
+++ b/tools/table_annovar/table_annovar.xml
@@ -52,7 +52,7 @@
         ## Post-processing: process annotated table to remove "NA" strings from numerical columns if
         ## tabular. Copy to output.
         #if str($out_format) == "tabular":
-            ; cat output.${input.dbkey}_multianno.txt | python replace_NA.py > ${output}
+            ; cat output.${input.dbkey}_multianno.txt | python ${__tool_directory__}/replace_NA.py > ${output}
         #else:
             ; cp output.${input.dbkey}_multianno.vcf ${output}
         #end if
--- a/tools/table_annovar/table_annovar.xml
+++ b/tools/table_annovar/table_annovar.xml
@@ -9,7 +9,7 @@
     <command>
         ## Convert VCF to AV input format.
         #if str($out_format) == "tabular":
-            convert2annovar.pl -format vcf4 -includeinfo ${input} > input.avinput ;
+            \${SCRIPT_PATH}/convert2annovar.pl -format vcf4 -includeinfo ${input} > input.avinput ;
             #set tab_anno_input = "input.avinput"
         #else:
             #set tab_anno_input = $input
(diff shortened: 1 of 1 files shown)


matthdsm:
Please don't merge this yet, apparently this breaks the replace_NA.py script.

matthdsm:
I fixed the issue, everything seems to function properly now.

Merge away if you please!

bgruening:
`${__tool_directory__}` looks good to me. Why do you use SCRIPT_PATH? Because both files needs to be installed manually, isn't it? They are not shipped with the repository. Or do I miss something?

matthdsm:
Hi Björn,

It's indeed as you expect. Annovar has the same issues as GATK when installing. The user needs to agree with a personal licence and manually install the scripts. In env.sh the user can then set the SCRIPT_PATH in the same fashion as the GATK_PATH is set in the GATK tool. That is also why I edited the README file in pull request #337.

I actually had some issues getting annovar to work in it's current form, since we're reluctant to edit the PATH for a single user, as everything is intertwined in our current setup. This is why I proposed these changes.

Matthias.

bgruening:
Oo, forgot about this one. I'm :+1:  on this one.
@galaxyproject/core anyone else?

tools/table_annovar/table_annovar.xml
```diff
@@ -42,7 +42,7 @@
         #set operation = ','.join( $operation )
 
         ## Annotate variants.
-        table_annovar.pl ${tab_anno_input} ${__get_data_table_entry__('annovar_indexes', 'dbkey', $input.dbkey, 'path')} -protocol ${protocol} -operation ${operation} -nastring '.' -buildver ${input.dbkey} --outfile output
+        \${SCRIPT_PATH}/table_annovar.pl ${tab_anno_input} ${__get_data_table_entry__('annovar_indexes', 'dbkey', $input.dbkey, 'path')} -protocol ${protocol} -operation ${operation} -nastring '.' -buildver ${input.dbkey} --outfile output
```
mvdbeek: With this `\${SCRIPT_PATH}/ ` in place, users can't append the scripts to the path anymore, since when  ${SCRIPT_PATH} is not set, it will read `/table_annovar.pl`. You could change this to \${SCRIPT_PATH}table_annovar.pl and document that SCRIPT_PATH needs to include the trailing slash. But then again I'm not quite sure why the script path can't be appended to PATH ?!

matthdsm: I was trying to edit to tool to match the setup for GATK. I find this approach arguably better then messing around with the environmental values of the galaxy user. this is, of course, highly subjective.

mvdbeek: > Also, I'm not able to permanently edit the PATH variable, since this is managed from above to protect the other users from themselves.

You only need to append to the PATH in one of the env.sh files, or you append to the PATH when you start galaxy's handlers, this will not affect anything except for the current galaxy process.

matthdsm: and that's were I went wrong...
You can go ahead and close this pull request if you like.

Thanks for clearing this up.

mvdbeek: Sure, thanks for understanding!
Feel free to drop by https://webchat.freenode.net/?channels=#galaxyproject if you need help figuring this out (and if you see a way to help the documentation that's always welcome as well).

PR #344, opened 2016-03-05
advanced travis testing

bgruening:
Let's start some testing.


bgruening:
This PR is now ready for review. It will enable testing for bowtie and bwa and a few other tools as proof of concept. Moreover, it shares the same testing code with the IUC repo and my galaxytools repo.

peterjc:
Sounds like a great idea, with the reassurance that this approach is already in use elsewhere. At some point I'll probably adopt this for my own tool testing which isn't taking full advantage of planemo yet.

bgruening:
@peterjc I use the conda dependencies already on our production server and it is great for testing. Even if conda dependencies are not used widely it is great for testing and we should get as many tools under testing as possible. Feel free to merge :)

bgruening:
Anyone? Can we merge this?

mvdbeek:
Yes, looks good! :)

bgruening:
Awesome! We have now all 3 repos under testing! Now lets move on and get all tool updates under testing!

PR #345, opened 2016-03-11
Add sanitizer

bgruening:


bgruening:
Anyone?

blankenberg:
Should we bump the version, since we're letting more chars through?

bgruening:
More chars will not change the result of old tool runs, isn't it? Moreover, it will result in a new revision of all EMBOSS tools, because they are all in one repo, correct? I would vote for not increase but can do so :)

bgruening:
Anyone?

martenson:
lgtm, thanks @bgruening !

bgruening:
Thanks for the merge!

martenson:
uploaded to MTS

Issue #348, opened 2016-03-14
Bamtools filter doesn't allow for filtering out specific chromosomes. 


zfrenchee:
There are two BAM file filtering utilities that devteam maintains: `Filter SAM or BAM, output SAM or BAM` which is also called `samtools_filter2` I believe, and `bamtools_filter`. The latter always gives me this error:

```
terminate called after throwing an instance of 'std::out_of_range'
  what():  vector::_M_range_check
Aborted (core dumped)
[W::bam_hdr_read] EOF marker is absent. The input is probably truncated.
```

Which I would be prepared to believe, but the first of those two tools doesn't give me an error. 

What I'm trying to do is remove mitochondrial reads from a BAM file. `samtools_filter2` requires that I enumerate all the `chr` that I want to keep, while `bamtools_filter` lets me select ones that I want to throw away. The latter of those is much more convenient, since I want to keep _everything except_ chrM. 

Is there any chance that `bamtools_filter` is broken, and if so, could it be fixed? Otherwise, is there a way of specifying `chr`s that you want to throw away with 'samtools_filter2` that isn't documented? 

Thanks!

zfrenchee:
One way of solving this would be to add a -U option, the way `samtools view` does, to the `samtools_filter2` tool.

zfrenchee:
When I use [Bamtools filter](https://github.com/galaxyproject/tools-devteam/blob/0037a546e17ed2c32c9bf662aec292e31ac5e653/tool_collections/bamtools/bamtools_filter/bamtools-filter.xml) to filter out low quality reads, it doesnt give me the out of range error, but when I ask it to filter anything relating to chromosomes, it _does_. I think it might be broken. I see that it's a fairly new tool, at a low version.

zfrenchee:
Solved in #351

PR #349, opened 2016-03-15
Provide optionally negative selection

zfrenchee:
To filter out reads from a chromosome. See #348

### DIFF
--- a/tools/samtool_filter2/samtool_filter2.xml
+++ b/tools/samtool_filter2/samtool_filter2.xml
@@ -1,7 +1,7 @@
 <tool id="samtool_filter2" name="Filter SAM or BAM, output SAM or BAM" version="1.1.2">
   <description>files on FLAG MAPQ RG LN or by region</description>
   <requirements>
-    <requirement type="package" version="0.1.2">samtools</requirement>
+    <requirement type="package" version="1.2">samtools</requirement>
   </requirements>
   <stdio>
     <exit_code range="1:" />
--- a/tools/samtool_filter2/samtool_filter2.xml
+++ b/tools/samtool_filter2/samtool_filter2.xml
@@ -17,7 +17,7 @@
   #set $input = 'input.sam'
   ln -s "$input1" $input &&
 #end if
-samtools view $pos_or_neg_select "$output1" $header
+samtools view -o $possibly_select_inverse "$output1" $header
 
   #if $input1.datatype.file_ext == 'sam':
    -S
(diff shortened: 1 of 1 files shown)


tools/samtool_filter2/samtool_filter2.xml
```diff
@@ -115,6 +115,7 @@ samtools view -o "$output1" $header
     <param name="read_group" type="text" value="" label="Select alignments from Read Group"
            help="(-r) Requires headers in the input SAM or BAM, otherwise no alignments will be output"/>
     <param name="bed_file" type="data" format="bed" optional="true" label="Output alignments overlapping the regions in the BED FILE" help="(-L)"/>
+    <param name="pos_or_neg_select" checked="false" label="Select everything but the following:" type="boolean" truevalue="-U" falsevalue="-o" help="" />
```
bgruening: `-o` needs to stay even if you want to have -U, correct? So just use falsevalue=""
I would call this option "Enable inverse selection" of something like that, the current naming sounds wired because it is referring to "following" which can be a lot.

tools/samtool_filter2/samtool_filter2.xml
```diff
@@ -17,7 +17,7 @@
   #set $input = 'input.sam'
   ln -s "$input1" $input &&
 #end if
-samtools view -o "$output1" $header
+samtools view $pos_or_neg_select "$output1" $header
```
bgruening: `-o` needs to stay
Please increase the version number of the tool.

bgruening:
Thanks @zfrenchee for working on this!

zfrenchee:
Alright. Let me know what you think now!

tools/samtool_filter2/samtool_filter2.xml
```diff
@@ -17,7 +17,7 @@
   #set $input = 'input.sam'
   ln -s "$input1" $input &&
 #end if
-samtools view -o "$output1" $header
+samtools view -o $possibly_select_inverse "$output1" $header
```
bgruening: `-o` should be infront of the output

zfrenchee:
Am I supposed to specify the filename twice, i.e. -o "$output1" $possibly_select_inverse "$output1" ?

I thought the way I had it initially with either -o or -U was correct.

bgruening:
You are right, if you only want to have one output, but you also need to increase the version number of samtools. Do you know when the -U option was introduced? In samtools 1.2 it is included. So it should be save to bump the requirement to 1.2 in the tool and the tool_dependency file.

zfrenchee:
So I realized the way you're probably supposed to do this is -o "$output1" -U "$output2". I'm not totally sure it will work with just -U, but it very well might (and I would argue it should). Nevertheless, we can go with two outputs, but I might argue that the term"filter" implies that there should be one. What do you think?

bgruening:
One is probably good in this case, thanks!

zfrenchee:
Alright, I changed it back. What do you think?

nsoranzo:
@zfrenchee See https://github.com/galaxyproject/tools-devteam/pull/349#issuecomment-197375777 , samtools 0.1.18 does not have the `-U` option.

zfrenchee:
Woops sorry.

zfrenchee:
Why isn't this working?

zfrenchee:
Moved to #350

PR #351, opened 2016-03-16
Allow for inverse selection (filtering out)

zfrenchee:
Partially solves #348


zfrenchee:
How do people feel about this @bgruening ?

bgruening:
I feel good about it, because the tests are passing :) So at least this will not break anything.
Sorry for not answering your mail so fast.
Anyone else? @davebx?

davebx:
Looks solid to me.

bgruening:
@davebx could you please update this wrapper?

bgruening:
Thanks a lot @zfrenchee for working on this!

bgruening:
@davebx I'm not sure but @zfrenchee probably needs this in usegalaxy.org. Is there a chance to update this as well?

zfrenchee:
I'm actually running my own server!

bgruening:
Even better :)

davebx:
I've updated both sheds.

Thanks for the contribution, @zfrenchee

zfrenchee:
I don't think this actually worked, sadly. I'm trying to understand the [samtools source](https://github.com/samtools/samtools/blob/develop/sam_view.c) and although I'm not sure I think it might be the case that a `-U` flag alone won't do it.

zfrenchee:
See https://github.com/samtools/samtools/issues/551#issuecomment-204566822

PR #355, opened 2016-03-21
Update kraken tool_dependencies.xml

blankenberg:


bgruening:
We should symlink those files or reorganise them and use planemo demultiplex for it.
Thanks @blankenberg!

PR #357, opened 2016-03-21
Add suite for kraken 0.10.6

blankenberg:


suites/suite_kraken_0_10_6/repository_dependencies.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0"?>
+<repositories description="A suite of Galaxy utilities associated with version 0.10.6 of the Kraken tool">
+    <repository name="data_manager_build_kraken_database" owner="devteam" />
+    <repository name="kraken" owner="devteam" />
+    <repository name="kraken_filter" owner="devteam" />
```
bgruening: I'm always confused about these versioned suites. 
We need to include here revisions, correct? Othwise, we do not now (in github) which version was targeted. For example if I want to re-upload this one into my own TS in 1 year, I will get different results.

PR #360, opened 2016-03-29
added '-f' flag to soft link (ln) commands. 

Takadonet:
Reason being is that if the job was requeued with slurm scheduler because a cluster node has died, the job_working_directory may not be empty and ln command fails because the soft link already exist.

### DIFF
--- a/tools/freebayes/freebayes.xml
+++ b/tools/freebayes/freebayes.xml
@@ -11,22 +11,22 @@
     #set $reference_fasta_filename = "localref.fa"
     
     #if str( $reference_source.reference_source_selector ) == "history":
-        ln -s "${reference_source.ref_file}" "${reference_fasta_filename}" &amp;&amp;
+        ln -f -s "${reference_source.ref_file}" "${reference_fasta_filename}" &amp;&amp;
         samtools faidx "${reference_fasta_filename}" 2&gt;&amp;1 || echo "Error running samtools faidx for FreeBayes" &gt;&amp;2 &amp;&amp;
     #else:
         #set $reference_fasta_filename = str( $reference_source.ref_file.fields.path )
     #end if
     
     #for $bam_count, $input_bam in enumerate( $reference_source.input_bams ):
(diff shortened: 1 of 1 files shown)


martenson:
this feels like something that should not be treated on the tool level but I do not know where else to put it :/

Takadonet:
I agree but it's the easiest solution to our problem. We have very large workflow runs (1k jobs+) that cannot be  recovered because a few jobs died. Frustrating for our users to re-submit.

Planning to do a few more pull request to add that '-f' flag. If you are planning to accept this one.

bgruening:
@martenson why does a new job get the same working directory? Is this a bug in the cleanup step?

Takadonet:
Its not a clean up bug.  When a node dies on our slurm cluster,  the jobs
on it will keep report they are running.  Slurm has ability to requeue the
job to another node with same job id.  Since its the same job id,  galaxy
never find out that it was resubmited.

bgruening:
@Takadonet there were a few SLURM bugs fixed recently. Is this still a problem for you? If so I will merge this. Sorry for taking such a long time.

nsoranzo:
:+1: Looks safe.

ericenns:
So I think the better solution to this problem is to auto clean the working directory in the job script sent to the cluster. This is still a problem for us, it isn't a slurm problem. The reason this happens is in our cluster jobs can get re queued if a higher priority job comes in. Then tools fail because they already ran for a bit and have partial output.

bgruening:
@ericenns I tend to agree! Maybe this issue should be moved to the galaxy repo. In the meantime I'm open to fix it here. @Takadonet ?

Takadonet:
I agree.  Please move to main repo

bgruening:
Hi,

we worked a little bit on freebayes in the meantime. I just added it here: https://github.com/galaxyproject/tools-devteam/commit/ec3aa49a9d65ed3b69836c357e8a0278cd034a75

Thanks!

PR #364, opened 2016-04-07
Update vcftools_merge.xml

ChristopheH:
The wrapper is now using the name of vcf input files as column header instead of the number of the file in the history. 
It also deals with the space and special characters in the files to replace them with "_".
Thanks to @nsoranzo and @jmchilton for helping.

### DIFF
--- a/tool_collections/vcftools/vcftools_merge/vcftools_merge.xml
+++ b/tool_collections/vcftools/vcftools_merge/vcftools_merge.xml
@@ -7,25 +7,28 @@
     </requirements>
 
     <command>
+    <![CDATA[
         ## Preprocessing for each dataset.
+        #import re
         #set dataset_names = []
         #for $i, $input in enumerate( $input_files ):
+            #set input_name = re.sub('[^\w\-_]', '_', $input.element_identifier)
             ## Sort file.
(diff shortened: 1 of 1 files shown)


tool_collections/vcftools/vcftools_merge/vcftools_merge.xml
```diff
@@ -7,25 +7,28 @@
...
-            tabix -p vcf ${i}.vcf.sorted.gz ;
+            tabix -p vcf '${input_name}.vcf.sorted.gz' &&
 
-            #silent dataset_names.append( str($i) + '.vcf.sorted.gz' )
+            #silent dataset_names.append( "'%s.vcf.sorted.gz'" % $input_name )
         #end for
 
         ## Merge.
         vcf-merge
         #echo ' '.join( dataset_names ) # > ${output}
```
nsoranzo: Quoting also `${output}` would be a plus.

nsoranzo:
@nekrut Is this change fine with you?

jmchilton:
If this tool is used with collections of vcf files - the collection identifier information will now be preserved in the output file - this is great. For normal datasets - I don't know if using the previous index or a name is better - but at very least this new modality should be an option I think.

nekrut:
👍

Issue #365, opened 2016-04-14
Tophat has a mismatch between default params and command-line triggering failures


jennaj:
**Issue**

This option Coverage Search (under full params) is marked as "No" by default, yet is not included in the command-line. This causes certain jobs to fail (likely for resource) but the error output is not captured as exceeding memory or wall-time. **UPDATE: THIS IS A GUESS, NOT CONFIRMED. The root cause of the problem is still an unknown.**

**UI setting**

<img width="700" alt="tophat_default_coverage_search" src="https://cloud.githubusercontent.com/assets/8999735/14537112/787bd91e-022a-11e6-89b4-b213ec3a5023.png">

**Command-line**

Note that it is missing
> --no-coverage-search

Command-line from job info page in UI

> Job Command-Line:	tophat2 --num-threads ${GALAXY_SLOTS:-4} -r 250 --mate-std-dev=20 --no-discordant /galaxy/data/mm10/bowtie2_index/mm10 "/galaxy-repl/main/files/014/976/dataset_14976631.dat" "/galaxy-repl/main/files/014/976/dataset_14976632.dat"

**Example Error**

Error for this type of job is always a trace dump and ending at the same point .. where junctions are written. I assume because there are way too many. Not all jobs fail for this .. tends to be linked with certain data combinations (fragmented genomes and/or high coverage/targetted region fastq seq reads = both tend to produce more discovered "junctions")

> job stderr:
> Fatal error: Tool execution failed
> 
> [2016-03-28 07:07:14] Beginning TopHat run (v2.0.14)
> -----------------------------------------------
> [2016-03-28 07:07:14] Checking for Bowtie
>                   Bowtie version:        2.2.5.0
> [2016-03-28 07:07:15] Checking for Bowtie index files (genome)..
> [2016-03-28 07:07:15] Checking for reference FASTA file
> [2016-03-28 07:07:15] Generating SAM header for /galaxy/data/mm10/bowtie2_index/mm10
> [2016-03-28 07:07:21] Preparing reads
>          left reads: min. length=75, max. length=75, 177927354 kept reads (23727 discarded)
>         right reads: min. length=75, max. length=75, 175340316 kept reads (2610765 discarded)
> [2016-03-28 08:40:24] Mapping left_kept_reads to genome mm10 with Bowtie2
> [2016-03-28 11:16:27] Mapping left_kept_reads_seg1 to genome mm10 with Bowtie2 (1/3)
[...]

PR #367, opened 2016-04-25
Add auto option to coverage_search conditional in TopHat

nsoranzo:
See #365 .

Ping @jennaj.

### DIFF
--- a/.tt_blacklist
+++ b/.tt_blacklist
@@ -52,7 +52,6 @@ tools/canonical_correlation_analysis
 tools/vcf_intersect
 tools/dwt_var_perfeature
 tools/best_regression_subsets
-tools/tophat2
 tools/plot_from_lda
 tools/count_gff_features
 tools/dwt_cor_avb_all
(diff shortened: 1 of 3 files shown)


jennaj:
+1 (but I don't really get to vote officially, this is just for liking the improved modeling of params  :) )

martenson:
@jennaj of course you get to vote! 
https://github.com/galaxyproject/galaxy/blob/dev/doc/source/project/organization.rst#members

bgruening:
Do we want to update to tophat 2.1.0 and bowtie 2.2.8? This would enable conda testing as well.

nsoranzo:
I can update to tophat 2.1.0, but there's no package_bowtie_2_2_8 yet.

nsoranzo:
@bgruening I've updated the wrapper to use TopHat 2.1.0 and removed the tool from the blacklist. There should be no need to use Bowtie 2.2.8 since 2.2.5 is already in Bioconda, but its build.sh seems broken (binaries are not moved to `bin/`). Compare:
https://github.com/bioconda/bioconda-recipes/blob/master/recipes/bowtie2/2.2.5/build.sh
https://github.com/bioconda/bioconda-recipes/blob/master/recipes/bowtie2/build.sh

Can you fix that?

nsoranzo:
I've fixed Bowtie2 2.2.5 in Bioconda, now tests pass. Merge?

bgruening:
Indeed thanks Nicola!

PR #371, opened 2016-05-03
Explicitly call python in command instead of using the depredated interpreter attribute

nsoranzo:
@natefoo on IRC:
> that reminds me, one of the tools that couldn't run on jetstream was due to its use of interpreter (which pulsar doesn't handle) 
> that tool (when I figure out which one it is) should be updated to use `$__tool_directory__`
it was one of the cuff* 
> cuffdiff maybe

This should fix all cufflinks tools.


bgruening:
Thanks Nicola!

jmchilton:
Thanks for the update. The IRC comment is exactly wrong though - he interpreter attribute lets Pulsar know what to transfer - the tool directory attribute is so unstructured that Pulsar doesn't know to transfer the file. Pulsar maintainer John is angry that tool framework maintainer John has unilaterally declared interpreter deprecated - but tool framework maintainer John had his reasons.

PR #372, opened 2016-05-22
Test more tools by default

bgruening:
Need to trigger a test here. But these repos should be ready for testing as they do not have dependency or only bx-python.

### DIFF
--- a/.tt_blacklist
+++ b/.tt_blacklist
@@ -33,7 +33,6 @@ tools/lda_analysis
 tools/ctd_batch
 tools/generate_pc_lda_matrix
 tools/convert_solid_color2nuc
-tools/tabular_to_fasta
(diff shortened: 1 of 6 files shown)


bgruening:
These very basic and important tools are ending up in so many Galaxy installations it would be good to ensure they are tested and conda-ready.

Same for https://github.com/galaxyproject/tools-devteam/tree/master/tool_collections/fastx_toolkit or https://github.com/galaxyproject/tools-devteam/tree/master/tool_collections/galaxy_sequence_utils

Help by converting them and reviewing is very much appreciated. I would like to have this in a solid state for GCC.

Thanks!

P.S. We could also think about adding python2.7 as a dependency. Or do we want to port the code at a given point?

PR #381, opened 2016-07-27
Adding "supplementary reads" filter to samtools filter

odoland:
There's a samflag 0x0800 for supplementary reads that I'd like to use on galaxy. Thanks!!


nsoranzo:
Thanks a lot @odoland!

Issue #391, opened 2016-09-15
VCF-VCFintersect needs to be updated in the Main Tool Shed and Galaxy Test/Main servers


jennaj:
Update tool shed then update test/main to correct usage issues (tool always errors).

Error

> The Galaxy framework encountered the following error while attempting to run the tool:
> Traceback (most recent call last):
>   File "/cvmfs/main.galaxyproject.org/galaxy/lib/galaxy/jobs/runners/__init__.py", line 170, in prepare_job
>     job_wrapper.prepare()
>   File "/cvmfs/main.galaxyproject.org/galaxy/lib/galaxy/jobs/__init__.py", line 897, in prepare
>     self.command_line, self.extra_filenames, self.environment_variables = tool_evaluator.build()
>   File "/cvmfs/main.galaxyproject.org/galaxy/lib/galaxy/tools/evaluation.py", line 445, in build
>     raise e
> ValueError: invalid literal for int() with base 10: 'use_adv_controls'
> Tool execution generated the following error message:
> invalid literal for int() with base 10: 'use_adv_controls'

Github master: https://github.com/galaxyproject/tools-devteam/tree/master/tool_collections/vcflib/vcfvcfintersect

Tool Shed repo: https://toolshed.g2.bx.psu.edu/repository?repository_id=fde54b460b1ce65c

Change from July that addresses the "use_adv_controls" code problem: https://github.com/galaxyproject/tools-devteam/commit/4cb83fa3d4c009796d9b87557d17a89f0148889d

Report at Biostars: https://biostar.usegalaxy.org/p/19440/

martenson:
I pushed the whole updated Conda-only vcflib to the MTS

martenson:
@natefoo do I need to do something special to update all of these on Main given it needs Conda to install its dependency?

natefoo:
@martenson Can the suite be updated so I don't have to update each tool individually? Or would that not work for updates anyway?

martenson:
@natefoo here you go, install this one: https://toolshed.g2.bx.psu.edu/view/devteam/suite_vcflib_1_0_0_rc1/2cafd05a4a59

martenson:
@natefoo do not install yet, there is no Conda package for one of the dependencies yet

martenson:
@natefoo the suite should be fixed now and ready to install via Conda

natefoo:
Looks like they installed correctly, but the version string was bad for vcfsort:

```
[...]

PR #392, opened 2016-09-15
remove the incorrect tool_dependencies.xml

martenson:
as the TS package does not exist in this version anymore; we rely solely on Conda here from now on

@bgruening @mvdbeek


martenson:
xref https://github.com/galaxyproject/tools-devteam/issues/391

Issue #389, opened 2016-09-08
fastq_trimmer_by_quality missing import six


afgane:
When running installed[ fastq_trimmer_by_quality tool](https://github.com/galaxyproject/tools-devteam/tree/master/tools/fastq_trimmer_by_quality) (at latest revision 04a609b0fdf6), it reports 
```
Traceback (most recent call last):
  File "/opt/galaxy/shed_tools/toolshed.g2.bx.psu.edu/repos/devteam/fastq_trimmer_by_quality/04a609b0fdf6/fastq_trimmer_by_quality/fastq_trimmer_by_quality.py", line 3, in <module>
    from galaxy_utils.sequence.fastq import fastqReader, fastqWriter
  File "/opt/galaxy/galaxy-app/lib/galaxy_utils/sequence/fastq.py", line 7, in <module>
    from six import Iterator, string_types
ImportError: No module named six
```

martenson:
some findings:

* https://api.anaconda.org/search?name=galaxy_sequence_utils has only version 1.0.1 so conda can never have non-versionless resolution for this since tool requires 1.0.0
* there is no v1.0.1 in the home repo of the package... (https://github.com/galaxyproject/sequence_utils)

afgane:
This appears to be caused by the order of dependency resolvers that was setup and because fastq.py v1.0.1 is not available in Conda. As an immediate fix, setting the order of dependency resolvers in `dependency_resolvers_conf.xml` as follows fixed the issue:
```
<dependency_resolvers>
  <conda />
  <tool_shed_packages />
  <galaxy_packages />
  <conda versionless="true" />
  <galaxy_packages versionless="true" />
</dependency_resolvers>
```

martenson:
@afgane I would call that a workaround rather than solution. I think we still have a bug in either Galaxy/Conda handling identical imports or in understanding how `versionless` resultion works

martenson:
ping @jmchilton @bgruening @mvdbeek

jmchilton:
What was broken resolver configuration? Does the default work?

```
<dependency_resolvers>
  <tool_shed_packages />
  <galaxy_packages />
  <galaxy_packages versionless="true" />
  <conda />
  <conda versionless="true" />
</dependency_resolvers>
```

afgane:
@jmchilton It was
```
<dependency_resolvers>
  <conda />
  <conda versionless="true" />
  <tool_shed_packages />
  <galaxy_packages />
[...]

PR #394, opened 2016-09-21
Fixes the missing module 'six' error in issue #389

nturaga:
https://github.com/galaxyproject/tools-devteam/issues/389

This should definetely be tested by someone from iuc/devteam because
conda-env in each job is not importing the modules in galaxy's virtualenv
'.venv'.


martenson:
This is not a proper fix as the problem is the tool is importing from Galaxy codebase instead of the dependency package. The discussion continues at #389.

PR #395, opened 2016-09-26
Picard rnaseqmetrics + GTF

yhoogstrate:
The picard tool for RNA-Seq metrics makes use of a format referred to as refFlat. This is not a very commonly used format which caries more or less the same information as GTF/GFF files. Therefore this patch that allows to also make use of those formats.

Due to all the changes with respect to conda, I don't know if the test will pass.


bwlang:
I love this idea!

bgruening:
@yhoogstrate please ping me. I keep loosing track. This is great!
Do you have time to get picard under conda testing entirely and remove the tool_dependencies.xml file?
If not please increase the tool version and I will merge it.

Thanks!

bgruening:
@yhoogstrate can you please rebase?

bgruening:
Great stuff!

PR #397, opened 2016-10-03
Add missing macros include to vcfsort

natefoo:
This should take care of the version string issue encountered in #391.


bgruening:
Thanks @natefoo!

PR #398, opened 2016-10-03
fixed wrapper

medda:
Hi, we fixed the pileup_parser.xml file because the pileup_parser.pl script contains a deprecated function (isdigit) which in our perl version (5.20.2) gives a warning (Calling POSIX::isdigit() is deprecated at pileup_parser.pl line 40, <IN> line 1. So the Galaxy job fails.

### DIFF
--- a/.tt_blacklist
+++ b/.tt_blacklist
@@ -73,7 +73,6 @@ tools/compute_motif_frequencies_for_all_motifs
 tools/compute_motifs_frequency
 tools/rmapq
 tools/show_tail
-tools/pileup_parser
 tools/table_annovar
 tools/lastz_paired_reads
 tools/t_test_two_samples
--- a/tools/pileup_parser/.shed.yml
+++ b/tools/pileup_parser/.shed.yml
@@ -1,6 +1,7 @@
 categories:
 - SAM
 description: Filter pileup on coverage and SNPs
+homepage_url: https://github.com/galaxyproject/tools-devteam/
 long_description: |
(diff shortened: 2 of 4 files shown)


nsoranzo:
@medda Thanks for reporting the bug! Unfortunately I don't think that's the proper fix, can you try changing
```
isdigit $fields[ $cvrg_column ]
```
to
```
$fields[ $cvrg_column ] =~ qr/^[[:digit:]]+$/x
```
in `pileup_parser.pl` ?

bgruening:
@medda thanks for your contribution. I also think we should fix the perl code as @nsoranzo mentioned. In addition let's define a strict dependency on perl from the conda.

```
    <requirements>
        <requirement type="package" version="5.22.0">perl</requirement>
    </requirements>
```

medda:
Ok, now the tool works well.
thank you

nsoranzo:
@medda I suppose you closed the PR by mistake, right?

medda:
I thought I'd have to close after I reported that works

peterjc:
Right now no one else will benefit from the fix. We don't close the pull request until a fix has been made to the main branch repository (and the Tool Shed updated etc).

nsoranzo:
I've activated Conda testing and fixed some old linting problems, merging now. Thanks @medda for reporting that the fix is working!

medda:
Ok, thanks, can I close the pull request?

nsoranzo:
@medda It's been already closed by merging.

medda:
Ok, thanks

PR #404, opened 2016-10-26
Add --gene_selector flag when generating heatmap with gene list

jj-umn:
The --genes args are ignored by cummeRbund.R when generating a heatmap unless the --gene_selector argument is also included.


bgruening:
Can you please bump the version. Thanks JJ!

bgruening:
Thanks @jj-umn

PR #408, opened 2016-10-31
FASTQC: add ability to detect .gz file when fastq.gz file is uploaded…

dwightkuo:
This commit tries to detect whether an incoming file is gzipped rather than relying solely on the name. This can be a problem when loading or importing files which are .gzip format but loaded as a different format (e.g. fastq.gz loaded as fastq). This is a common use case when dealing with fastq.gz files. Note that in the case of a fastq.gz file that is not named as "fastq.gz" the "format" option for the python script needs to be removed since the program fastqc has no option to specify "fastq.gz" directly from the command line

### DIFF
--- a/tools/fastqc/rgFastQC.py
+++ b/tools/fastqc/rgFastQC.py
@@ -28,6 +28,7 @@ import glob
 import gzip
 import bz2
 import zipfile
+import mimetypes
 
 class FastQCRunner(object):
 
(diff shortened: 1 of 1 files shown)


bgruening:
Smart solution! Thanks a lot!
Could you add a minimal test case for this as well? One- Two lines of gzipped fastq?

PR #410, opened 2016-11-02
Fastqc addgz

dwightkuo:
added a test case using a fastq.gz file. Note that this is just the "1000gsample.fastq" file gzipped as "1000gsample.fastq.gz". File could not be added to repository sinze .gitignore prevents .gz files from being contributed


bgruening:
Can you try `git commit file.gz -f` to force a commit?

bgruening:
:+1: Thanks a lot @dwightkuo!

PR #413, opened 2016-11-07
Enhancements to xy_plot

shiltemann:
Want to use this tool for a training, made the following enhancements to make it a bit more robust/user-friendly:

- allow user to indicate if there is a header line present
- allow for NA values in the input file
- allow plotting of multiple columns of a file per series
- allow user to choose between pdf and png output


bgruening:
Looks good to me!

bgruening:
Thanks Saskia!

PR #402, opened 2016-10-16
vcftools-consensus from vcftools

pvanheus:
This provides the vcftools-consensus tool that merges a FASTA and VCF file to generate a new FASTA sequence derived from applying the variants in the VCF file to the FASTA reference sequence. 

I am contributing it here because of the vcftools collection in this tools-devteam repo but could equally add it to tools-iuc since there is nothing, per se, shared with the collection (macros are not used, for example).

PR #417, opened 2016-11-25
vcftools_consensus:  Add tool_data_table_conf and associated files

pvanheus:
Based on the error "This file requires an entry in the tool_data_table_conf.xml file. Upload a file named tool_data_table_conf.xml.sample to the repository that includes the required entry to correct this error." added tool_data_table_conf.xml.sample and associated files. This follows on PR #402.


nsoranzo:
Thanks @pvanheus!

PR #420, opened 2016-12-06
Escape element identifiers in picard

mvdbeek:
ping @jmchilton  @bgruening

### DIFF
--- a/tools/picard/picard_macros.xml
+++ b/tools/picard/picard_macros.xml
@@ -18,7 +18,7 @@
 
     <token name="@symlink_element_identifier@"><![CDATA[
     #import re
-    #set escaped_element_identifier = re.sub('[^\w\-_]', '_', str($inputFile.element_identifier))
+    #set escaped_element_identifier = re.sub('[^\w\-]', '_', str($inputFile.element_identifier))
     ln -f -s '$inputFile' '$escaped_element_identifier' &&
     ]]></token>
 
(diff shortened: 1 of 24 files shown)


bgruening:
John is this really needed? Why can this not be handled by Galaxy.

tools/picard/picard_macros.xml
```diff
@@ -17,7 +17,9 @@
     </xml>
 
     <token name="@symlink_element_identifier@"><![CDATA[
-    ln -f -s '$inputFile' '$inputFile.element_identifier' &&
+    #import re
+    #set escaped_element_identifier = re.sub('[^\w\-_]', '_', str($inputFile.element_identifier))
```
nsoranzo: The underscore should already be included in `\w`, see https://docs.python.org/2/howto/regex.html#matching-characters

jmchilton:
@bgruening If the framework does it - it might mess up analyses that are keyed on this correctly - it is like Excel guessing what datatype your input is and adjusting it. How it is escaped could depend on how it is used in the tool wrapper (quote types, configfile vs command, etc...). If you want to implement a ``$input.element_identifier_shell_escaped``` or something I'd be happy to do that.

mvdbeek:
> @bgruening If the framework does it - it might mess up analyses that are keyed on this correctly - it is like Excel guessing what datatype your input is and adjusting it. How it is escaped could depend on how it is used in the tool wrapper (quote types, configfile vs command, etc...). If you want to implement a ``$input.element_identifier_shell_escaped``` or something I'd be happy to do that.

In that particular case the framework should just do the (escaped) symlinking.

bgruening:
I think more framework support for these kind of stuff would be nice yes!

bgruening:
Thanks Marius!

PR #426, opened 2017-01-03
Bring bowtie2 up to current standards

mvdbeek:
upgrade to bowtie 2.2.8, upgrade samtools to 1.3.1 and add multi-threading when indexing.


bgruening:
Awesome!

PR #427, opened 2017-01-08
bowtie2 update

bgruening:


mvdbeek:
Nice! This one is supposed to be quite a bit faster in multithreading!

PR #434, opened 2017-01-16
Replace isinstance with is_of_type

nsoranzo:
xref. https://github.com/galaxyproject/galaxy/issues/3353

Also update a bit the affected tools to recent standard.


mvdbeek:
Looks good, but tests didn't run. I'll open and close to see if it's being picked up.

bgruening:
Since when is this in place? Should we increase the version number?

mvdbeek:
2014-08-12 -- since it doesn't change the output I would say no.

bgruening:
I was more worried about backwards compatibility but 2014 seems to be old enough to me! :)

bgruening:
Thanks  @nsoranzo and @mvdbeek !

PR #441, opened 2017-01-24
Update bwa for fastqsanger.gz support

mvdbeek:
This is very straightforward since bwa doesn't rely on the file extension to detect compression.


tools/bwa/bwa.xml
```diff
@@ -331,6 +331,15 @@
     <test>
       <param name="reference_source_selector" value="history" />
       <param name="ref_file" ftype="fasta" value="bwa-mem-mt-genome.fa"/>
+      <param name="input_type_selector" value="paired"/>
+      <param name="fastq_input1" ftype="fastqsanger.gz" value="bwa-mem-fastq1.fq.gz"/>
```
bgruening: If automatic sniffing will be disabled by default, you might want to explicitly give the ftype?

mvdbeek: it's already there.

mvdbeek:
thanks @yhoogstrate @bgruening :)

PR #412, opened 2016-11-07
Cufflinks conda

lecorguille:
The PR aims to prepare cufflinks tool_collection to be conda ready.

###Cufflinks
I had deps for lib/galaxy/datatypes/util/gff_util.py which is called by [cufflinks_wrapper.py ](https://github.com/lecorguille/tools-devteam/blob/e9cb7de8b3789ecf6710faff54c88bb771b28a2d/tool_collections/cufflinks/cufflinks/cufflinks_wrapper.py#L9)
(it was tricky to solve because it seems a Galaxy library (so usually run in the .venv) but it's actually launched within the conda environment)

###Cuffdiff
I had to change the package name to fit with conda nomenclature.
I can change easily the ToolShed tooldependency package_cummerbund_2_8_2 from IUC to fit also with the conda nomenclature!?
https://github.com/galaxyproject/tools-iuc/blob/master/packages/package_cummerbund_2_8_2/tool_dependencies.xml#L9

```xml
 <package name="cummeRbund" version="2.8.2">
```
To
```xml
 <package name="bioconductor-cummerbund" version="2.8.2">
```

That way, we will remained compatible with both resolvers.

PR #447, opened 2017-01-31
Remove <trackster_conf> elements

nsoranzo:
Also condafy VarScan and remove it from the blacklist.

xref. https://github.com/galaxyproject/galaxy/pull/3133

I didn't touch Cufflinks because it probably should be fixed in #412.

Ping @jgoecks.


mvdbeek:
thanks @nsoranzo

PR #449, opened 2017-02-03
Implement compressed file support for fastq_manipulation.

jmchilton:
Requires compressed io support in galaxy_sequence_utils from https://github.com/galaxyproject/sequence_utils/pull/11.

To test this tool with that unpublished support - you can use the script in sequence_utils for testing tools with a local copy of sequence_utils.

```
$ git clone https://github.com/galaxyproject/sequence_utils.git
$ cd sequence_utils
$ git checkout compressed_io
$ ./tests/planemo_test.bash /path/to/tools-devteam/tool_collections/galaxy_sequence_utils/fastq_manipulation/
```

This also requires you have a ``~/.planemo.yml`` file setup to target Galaxy at https://github.com/galaxyproject/galaxy/pull/3550.


mvdbeek:
We'll also need to update the xsd for the decompress attribute. ```/tmp/tmpvJxrZ3:402:0:ERROR:SCHEMASV:SCHEMAV_CVC_COMPLEX_TYPE_3_2_1: Element 'output', attribute 'decompress': The attribute 'decompress' is not allowed.```

tool_collections/galaxy_sequence_utils/fastq_manipulation/fastq_manipulation.xml
```diff
@@ -2,9 +2,9 @@
...
   <description>reads on various attributes</description>
   <requirements>
-    <requirement type="package" version="1.0.1">galaxy_sequence_utils</requirement>
+    <requirement type="package" version="1.1.0">galaxy_sequence_utils</requirement>
   </requirements>
-  <command>python $__tool_directory__/fastq_manipulation.py '$input_file' '$fastq_manipulation_file' '$output_file' '$output_file.files_path' '${input_file.extension[len( 'fastq' ):]}'</command>
+  <command>gx-fastq-manipulation '$input_file' '$fastq_manipulation_file' '$output_file' '$output_file.files_path' '${input_file.extension[len( 'fastq' ):]}'</command>
   <configfiles>
     <configfile name="fastq_manipulation_file">##create an importable module
 #import binascii
```
nsoranzo: The code in this `configfile` is not Python3 compatible (e.g. uses `string.maketrans()`). Not sure keeping this big block of code here is a good idea anyway.

bgruening:
ping @jmchilton

nsoranzo:
Wait @jmchilton, I've been working on this and will push to this branch later today.

tool_collections/galaxy_sequence_utils/fastq_manipulation/fastq_manipulation.xml
nsoranzo: @jmchilton I've fixed the problem with `string.maketrans` in Python3, but I still don't understand what is the purpose of using `binascii.unhexlify("${ binascii.hexlify(str(match_block['match_type']['match']['match_by'])) }")` instead of just `$match_block['match_type']['match']['match_by']` for all text params, any clue?

nsoranzo:
I've pushed my fixes, 2 remaining problems:
- https://github.com/galaxyproject/sequence_utils/pull/14 needs to be merged, then we have to do a new PyPI release, update the Bioconda package and the requirement in the tools
- need to fix `fastq_manipulation` under Python3, see https://github.com/galaxyproject/tools-devteam/pull/449#discussion_r137520154

tool_collections/galaxy_sequence_utils/fastq_manipulation/fastq_manipulation.xml
jmchilton: I don't have a clue... security related maybe?

jmchilton: If somehow this circumvents the *Wrapper crap Galaxy does - this could make sense I suppose. I thought the parameters were still wrapped when using in configfiles though - I could be wrong.

nsoranzo: Ping @blankenberg

blankenberg: @jmchilton is correct with his thinking here. The sanitizer has been turned off on this tool, so we avoid issues by hexlify user text values.

nsoranzo: Thanks @blankenberg! Do you think that's still the best way to implement this?

blankenberg: Good question @nsoranzo. This fastq manipulation tool is very Galaxy-centric in its design. To allow the full complement of user text input values, while preventing security issues it is probably hard to beat -- properly quoting and unquoting strings safely can be difficult and this avoids any complications with ease. 

Rewriting the tool to no longer use configfiles, but instead make use of long command-line option sets may make it more generally useful outside of Galaxy, but isn't something I'd have any time to explore currently.

nsoranzo: @jmchilton @blankenberg I've fixed this by adding a `.decode()` after each `binascii.hexlify()`.

nsoranzo:
Ready to review/merge!

bgruening:
LGTM! Thanks @jmchilton and @nsoranzo. These are important tools. We should move them also to IUC.

PR #453, opened 2017-02-18
Remove tool_dependencies.xml for cufflinks, add parse_gff_attributes …

mvdbeek:
…and gff_attributes_to_str to avoid importing from galaxy's datatypes.

This will allow us to remove cufflinks >= 2.2.1.1 from the list of tools that need to preserve the python environment and simplify our life in the long run. I also removed the code for checking the version, I think that's better done using the `<version_command>` tag


mvdbeek:
I'm also thinking about removing the NamedTemporaryFile stuff. I think we should just write into the job working directory. Opinions?

mvdbeek:
This is failing with
```
2017-02-18 20:12:02,717 DEBUG [galaxy.tools.deps.conda_util] Executing command: /home/travis/conda/bin/conda create -y --name mulled-v1-bb234c7ebff47bed8a89c940146c1ee7031868d43c00a15be87abdbaa8f26dbd cufflinks=2.2.1 bioconductor-cummerbund=2.8.2
PaddingError: Placeholder of length '80' too short in package bioconda::r-fastcluster-1.1.20-r3.3.1_0.
The package must be rebuilt with conda-build > 2.0.
2017-02-18 20:12:12,170 DEBUG [galaxy.tools.deps.resolvers.conda] Removing failed conda install of [CondaTarget[cufflinks,version=2.2.1], CondaTarget[bioconductor-cummerbund,version=2.8.2]]
```

mvdbeek:
That's not a new error:
https://travis-ci.org/galaxyproject/tools-devteam/jobs/199322959#L1881.
I guess it passed because the previous test was not building a mulled environment, and so it was just removing cummerbund, and the presence or absence of cummerbund did not trigger a test failure. @bgruening any idea how to tackle the issue? I suppose something in the dependency chain of r-fastcluster-1.1.20-r3.3.1_0 needs to be rebuild ...

bgruening:
@mvdbeek I might have fixed this error, but there is now a new one :(

mvdbeek:
Yeah, this doesn't look like a conda error anymore. I'll have a look if this is just the more stringent `detect_errors="aggressive"` or whether the input bam is corrupt.

mvdbeek:
OK, so that's just a warning that is being output by samtools. The tool in question is cuffdiff, the correct output is generated and adding detect_errors="aggressive" solves the issue locally. Fingers crossed.

tool_collections/cufflinks/cuffdiff/cuffdiff_wrapper.xml
```diff
@@ -287,7 +286,7 @@
                 <param name="do_normalization" value="No" />
                 <param name="multiread_correct" value="No"/>
                 <param name="sAdditional" value="No"/>
-                <output name="output_cummerbund" ftype="sqlite" file="cuffdiff_out.sqlite" compare="sim_size" />
+                <output name="output_cummerbund" ftype="sqlite" file="cuffdiff_out.sqlite" compare="sim_size" delta="20000" />
```
nsoranzo: Maybe you don't need to increase the delta if you update cuffdiff_out.sqlite, which I think it's fair since bioconductor-cummerbund version has been increased.

tool_collections/cufflinks/cuff_macros.xml
```diff
@@ -1,5 +1,5 @@
 <macros>
-  <token name="@VERSION@">2.2.1</token>
+  <token name="@VERSION@">2.2.1.2</token>
```
nsoranzo: I think I prefer the previous approach were this is equal to the Cufflinks version and each tool updates its Galaxy version only when needed.

nsoranzo:
> I'm also thinking about removing the NamedTemporaryFile stuff. I think we should just write into the job working directory. Opinions?

In most cases these temporary files are created in the PWD already, just the name is random.

mvdbeek:
> In most cases these temporary files are created in the PWD already, just the name is random.

What's the mechanism there? Aren't they created in whatever $TMP points to?

nsoranzo:
But e.g. in cuffcompare_wrapper.py it is called with 

tempfile.NamedTemporaryFile(dir=".")

bgruening:
Nice, seems to work!
@mvdbeek I just rebuild the r-fastcluster library. Thanks all!

PR #459, opened 2017-04-10
Update cuff_macros.xml

lecorguille:
For some users, if they are not used to the term "Dataset collection", it can be confusing for example with `Cuffmerge`.
- They expect the same behaviour as `Concatenate datasets` where you add the datasets with a repeat but the input don't show any dataset because it want a dataset collection.
- Moreover, they think that if they select multiple dataset in the first input `multiple="true"`, they think that the tool will be launched in batch mode. Indeed, the `This is a batch mode input field. Separate jobs will be triggered for each dataset selection.` message only appear if you are in a case of batch mode. It should be a message for the `multiple="true"` which say `This is not a batch mode` 😕 

I don't know exactly which term but the idea is to use the term datatset collection somewhere in the label or in the help.

### DIFF
--- a/tool_collections/cufflinks/cuff_macros.xml
+++ b/tool_collections/cufflinks/cuff_macros.xml
@@ -66,7 +66,7 @@
   </token>
   <xml name="cufflinks_gtf_inputs">
     <param format="gtf" name="inputs" type="data" label="GTF file(s) produced by Cufflinks" help="" multiple="true" />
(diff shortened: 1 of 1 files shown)


bgruening:
mh ... ping @galaxyproject/iuc and @nekrut for the UX part.

bgruening:
Let's merge this :) Sorry for taking so long.

PR #463, opened 2017-04-26
Use requirements that exist in bioconda

davebx:


tool_collections/taxonomy/kraken2tax/kraken2tax.xml
```diff
@@ -1,8 +1,8 @@
 <tool id="Kraken2Tax" name="Convert Kraken" version="1.1">
```
nsoranzo: Can you increase the tool version?

nsoranzo:
We can also remove `tool_dependencies.xml` at this point, I think.

davebx:
Freedom eyes strikes again.

bgruening:
Thanks @davebx and @nsoranzo!

bgruening:
@davebx more of these!!!

nsoranzo:
I just noticed that this tool is not properly tested because `tool_collections/taxonomy/kraken2tax` is still in `.tt_blacklist` . Can you open an additional PR which removes this line and also does a small change to the tool to trigger the planemo tests?

PR #464, opened 2017-04-28
Fix Python3 compatibility for bowtie_wrappers and bowtie_color_wrappers

nsoranzo:
Also:
- use `with` statement
- flake8 fixes.

Without these, `bowtie_wrappers` will not work with Conda dependencies if using Miniconda3.
Reported by @lparsons.


bgruening:
Thanks @nsoranzo!

PR #467, opened 2017-05-09
Updated bamtools filter

davebx:
Highlights:

 - Can now consume a list of datasets, produces an identically structured list of datasets.
 - Produces multiple output files if given multiple input files.
 - Generates the json for bamtools in a more structured manner.
 - Updated tests to correspond.


mvdbeek:
> Can now consume a list of datasets, produces an identically structured list of datasets.
> Produces multiple output files if given multiple input files.

it was already able to do all of this by mapping over a collection, and now we're loosing the intrinsic parallelization of collections :(. Is there a reason for making this tool collection aware?

davebx:
@mvdbeek The old behavior was to consume n datasets, either in a list or individually, and produce a single merged filtered bam file. I was asked to change that behavior to make its output datasets correspond to the input datasets.

nekrut:
Yes, the previous behavior was wrong. It should not merge collections. This fixes that.

tool_collections/bamtools/bamtools_filter/bamtools-filter.xml
```diff
@@ -1,322 +1,346 @@
...
+                        <valid initial="string.printable"/>
+                    </sanitizer>
+                </param>
+            </when>
+        </conditional>
+    </inputs>
+    <outputs>
+        <data format="txt" name="out_file2" />
+        <data name="out_file1" format="bam" label="${tool.name} on ${on_string}">
+            <discover_datasets assign_primary_output="True" directory="outputs" pattern="filtered_(?P&lt;designation&gt;.+)\.bam" ext="bam" visible="true" />
```
nsoranzo: Planemo says: Element 'discover_datasets', attribute 'assign_primary_output': 'True' is not a valid value of the atomic type 'xs:boolean'.

I think you should use lowercase `true`.

jmchilton:
I'm with @mvdbeek - this isn't the correct thing to do at all.

It would be a lot better to wrap the input into a conditional - select at the beginning if you want to select a single dataset to filter or many. Default it to the single version and Galaxy would take care of all of these details - the other reduction variant would also be available if you swapped the conditional.

It is a limitation of Galaxy that you can't "map" a collection over a multiple data parameter - but the answer isn't to do the mapping yourself.

nekrut:
Aha, I have not thought of it this way. Indeed, this makes total sense

jmchilton:
> Yes, the previous behavior was wrong. It should not merge collections. This fixes that.

If it really should never merge collections - just make the multiple data parameter a single data parameter and Galaxy takes care of these details. If it should conditionally merge collections - implement the conditional like I described or make two tools.

jmchilton:
@nekrut Great - glad it makes sense now!

davebx:
Almost done updating to behave as @jmchilton is suggesting. Is there an example somewhere of testing a non-multiple input with a collection?

jmchilton:
@davebx One cannot write tests for this - Galaxy pretty much takes care of all the details related to rebuilding and mapping over the collection anyway though - so if it works for a single data input it should work with a collection input.

davebx:
@nsoranzo, @mvdbeek, any additional obstacles to getting this merged?

tool_collections/bamtools/bamtools_filter/bamtools-filter.xml
```diff
@@ -1,322 +1,299 @@
...
+    #if str( $rule_configuration.rules_selector ) == "True":
+        #set $config[ 'rule' ] = $rule_configuration.rules
+    #end if
+#end if
+$json.dumps( $config, indent=4 )
+]]>
+        </configfile>
+    </configfiles>
+    <inputs>
+        <param name="input_bams" type="data" format="bam" label="BAM dataset(s) to filter" />
```
nsoranzo: I think we should remove `(s)` in the label and maybe also change the name to `input_bam`.

tool_collections/bamtools/bamtools_filter/bamtools-filter.xml
```diff
@@ -1,322 +1,299 @@
...
+    <description>BAM datasets on a variety of attributes</description>
+    <requirements>
+        <requirement type="package" version="2.4.0">bamtools</requirement>
+    </requirements>
+    <command>
+        <![CDATA[
+            cat $script_file > $out_file2;
+            ln -s "${input_bams}" "localbam.bam" &&
+            ln -s "${input_bams.metadata.bam_index}" "localbam.bam.bai" &&
+            bamtools filter -script $script_file -in "localbam.bam" -out "$out_file1"
```
nsoranzo: Use single quotes around `${input_bams}`, `${input_bams.metadata.bam_index}` and `$out_file1`.
Add single quotes around `$script_file` (twice) and `$out_file2`.
You can also remove quotes around `localbam.bam` and `localbam.bam.bai`.

tool_collections/bamtools/bamtools_filter/bamtools-filter.xml
```diff
@@ -1,322 +1,299 @@
...
-    </citations>
-</tool>
+<tool id="bamFilter" name="Filter" version="2.4.1">
+    <description>BAM datasets on a variety of attributes</description>
+    <requirements>
+        <requirement type="package" version="2.4.0">bamtools</requirement>
+    </requirements>
+    <command>
+        <![CDATA[
+            cat $script_file > $out_file2;
```
nsoranzo:
```
cp '$script_file' '$out_file2' &&
```

tool_collections/bamtools/bamtools_filter/bamtools-filter.xml
```diff
@@ -1,322 +1,299 @@
...
+            ln -s '${input_bam}' localbam.bam &&
+            ln -s '${input_bam.metadata.bam_index}' localbam.bam.bai &&
+            bamtools filter -script $script_file -in localbam.bam -out '$out_file1'
```
nsoranzo: Single quotes around `$script_file`

bgruening:
ping @mvdbeek

tool_collections/bamtools/bamtools_filter/bamtools-filter.xml
```diff
@@ -1,322 +1,299 @@
...
+#if $conditions:
+    #set $config[ 'filters' ] = []
+    #for $i, $condition in enumerate( $conditions, start=1 ):
+        #set $filter = dict( id=str( $i ) )
+        #for $j, $s in enumerate( $condition.filters, start=1 ):
+            #set $filter[ $str( s[ 'bam_property' ][ 'bam_property_selector' ] ) ] = $str( s[ 'bam_property' ][ 'bam_property_value' ] )
+        #end for
+        $(config[ 'filters' ].append( $filter ))
+    #end for
+    #if str( $rule_configuration.rules_selector ) == "True":
```
mvdbeek: I think this is not going to work, since the truevalue is "true", and that is a string, not a bool (You can use the truevalue to pass commandline arguments). You'd have to change either the truevalue to `"true"` or this to `== "true"`
```
<param name="rules_selector" type="boolean" truevalue="true" falsevalue="false" label="Would you like to set rules?" help="Allows complex logical constructs. See Example 4 below." />
```

mvdbeek:
Thanks a lot @davebx !

PR #468, opened 2017-05-09
bamtools-split enable output as collection for ref and tag

jj-umn:
Add selector for -reference and -tag to have outputs in a dataset
collection.  This allows a HISAT bam to be split by chromosome and the
collection can be processed in parallel, e.g. hisat -> bamtools_split
=> bcftools_mpileup => bcftools_call => bcftools_concat -> snpEff.


jj-umn:
@jmchilton  John, since you just reviewed bamtools filter, could I ask you to take a quick look at this?  In particular the output_collection.  I'm using bamtools split to get a bam per chromosome, and keeping the order in the collection the same as the sort order of the input bam.  I implemented bcftools mpileup with a conditional for either single or multiple input bam.  This allows a HISAT bam to be split by chromosome and the collection can be processed in parallel, e.g. hisat -> bamtools_split
=> bcftools_mpileup => bcftools_call => bcftools_concat -> snpEff.

tool_collections/bamtools/bamtools_split/bamtools-split.xml
```diff
@@ -1,108 +1,146 @@
...
-    <citations>
-        <citation type="doi">10.1093/bioinformatics/btr174</citation>
-    </citations>
-</tool>
+<tool id="bamSplit" name="Split" version="2.4.1">
+    <description>BAM datasets on variety of attributes</description>
+    <macros>
+        <xml name="macro_output_type">
+            <param name="output_type" type="select" label="output as">
+                <option value="history_items">history items</option>
```
jmchilton: I'd display these to the user as "datasets" and "dataset list".

tool_collections/bamtools/bamtools_split/bamtools-split.xml
```diff
@@ -1,108 +1,146 @@
...
+                           if [ -e \$SN ]; 
+                             then FN=`printf "split_bam%05d%s.%s.bam" \$((I)) "$name" "\$i"`; 
+                             mv \$SN \$FN;
+                           fi;
+                        done)
+            #end if
+        ]]>
+    </command>
+    <inputs>
+        <param name="input_bams" type="data" format="bam" label="BAM dataset(s) to filter" min="1" multiple="True"/>
```
jmchilton: I'd consider making this just a single dataset input - this way if the users come with a dataset list the tool will produce a nested list naturally. I assume that would be the most common use case - ping @nekrut. If you still want to filter multiple individual bams into a flat list - I'd guess I'd make this a conditional but then you'd have to explain that operation and that wouldn't be easy.

tool_collections/bamtools/bamtools_split/bamtools-split.xml
```diff
@@ -1,108 +1,146 @@
...
+            <when value="-tag">
+                <param name="tag_name" type="text" value="NM" label="Enter tag name here" help="For example, to split on NM tag enter &quot;NM&quot;"/>
+                <expand macro="macro_output_type" />
+            </when>
+        </conditional>
+    </inputs>
+    <outputs>
+        <data format="txt" name="report" label="BAMSplitter Run" hidden="true">
+            <discover_datasets pattern="split_bam\.(?P&lt;designation&gt;.+)\.bam" ext="bam" visible="true"/>
+            <filter>analysis_type['analysis_type_selector'] in ('-mapped','-paired') or analysis_type['output_type'] != 'dataset_collection'</filter>
```
jmchilton: Discovered datsets just make little sense in the case of paired and mapping status - it should be two separate outputs. That seems like a useful operation in workflows for instance but we cannot use this tool in workflows in these cases.

I'd split this one tool into three Galaxy tools with fixed output types:

- Bam Split on reference or tag - always produces a dataset list.
- Bam Split paired - always produces a dataset pair.
- Bam Split on mapping status - always produces two datasets.

Ping @nekrut

jj-umn:
@jmchilton @nekrut I was trying to preserve backward compatibility.  But I agree that it makes more sense to makes more sense to have 3 separate tools that perform in a more intuitive manner.  I'll proceed with that.  Thanks.

jmchilton:
@jj-umn Great - thanks! I'd say especially given that this tool isn't written in a way that it is useful in workflows I think down prioritizing backward compatibility over getting it right is a good choice. Thanks again!

jj-umn:
@jmchilton I left the original tool as is, and added specific tools for each split type with outputs as you suggested.   Only remaining question is whether this repository should have .shed.yml declare "auto_tool_repositories" so that one could load an individual tool as needed.

tools/@
```diff
@@ -0,0 +1,290 @@
+#!/usr/bin/env python
+
```
bgruening: This file name looks wired.

tool_collections/vcflib/vcf2fasta/tool_dependencies.xml
```diff
@@ -0,0 +1 @@
+../tool_dependencies.xml
```
bgruening: Do we still need this file?

bgruening:
+1 for the auto-tool-repositories. I think in this PR are some file changes that do not belong here.
Thanks JJ.

tools/@
bgruening: ping @jj-umn I think you can remove this file

bgruening:
@jj-umn there are sill some files in this PR which are wrong. Like tool_collections/vcflib/vcf2fasta/test-data/vcf2fa_groupA_phiX174:0.fasta

davebx:
Implemented in galaxyproject/tools-iuc#1359

PR #470, opened 2017-05-19
migrate bowtie2 to legacy

martenson:
counterpart of https://github.com/galaxyproject/tools-iuc/pull/1300/


martenson:
bowtie is now under the IUC's wings

PR #471, opened 2017-05-22
Adopt and update Guru's heatmap tool.

davebx:
The first commit is the unmodified heatmap tool, the second updates it to follow better practices.


bgruening:
@davebx I think we should stop to add tools to tools-devteam. Can you please add it to IUC?

davebx:
Certainly. Adding to IUC.

PR #475, opened 2017-07-06
Add support for Python3 and condafy fastq_groomer

### LINKED ISSUE #474: Fastq Groomer sometimes is installed with Python 3 in Conda
But the wrapper script doesn't support it. At least this is what I suspect is happening.

I have an e-mail from a researcher at the NIH which states:

> One of the tools I use is the Fastq groomer which I have installed through the tool shed. I have searched a lot for the source of this error and I can’t find any solution. Every time I run this tool, I get this:

``` 
File “…/shed_tools/tooldshed.g2.bx.psu.edu/repos/devteam/fastq_groomer/e4d28c94242/fastq_groomer/fastq_groomer.py”, line 29
Print “Groomed %i %s reads into %s.” % (read_count + 1, input_type, output_type)
 
SyntaxError: Invalid syntax
 ```
 
> I am currently running python 2.7.8 and I did not find a: from __future__ import print_function anywhere so I am confused as why this error continually appears.

I suspect finishing up and merging https://github.com/galaxyproject/tools-devteam/pull/449 should fix this problem since that uses an update galaxy sequence utils with Python 3 support.

nsoranzo:
- Fix #474
- Remove deprecated `interpreter` attribute of `<command>`
- Fix indentation
- Add CDATA

This is going to conflict with #449, but I think it's worth to fix this tool ASAP.

Ping @jmchilton.

### DIFF
--- a/tool_collections/galaxy_sequence_utils/fastq_groomer/fastq_groomer.py
+++ b/tool_collections/galaxy_sequence_utils/fastq_groomer/fastq_groomer.py
@@ -1,6 +1,15 @@
-#Dan Blankenberg
+# Dan Blankenberg
+from __future__ import print_function
+
 import sys
-from galaxy_utils.sequence.fastq import fastqReader, fastqVerboseErrorReader, fastqAggregator, fastqWriter
+
+from galaxy_utils.sequence.fastq import (
+    fastqAggregator,
+    fastqReader,
+    fastqVerboseErrorReader,
+    fastqWriter
+)
+
 
 def main():
     input_filename = sys.argv[1]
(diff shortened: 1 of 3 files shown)


jmchilton:
I don't know - I feel like we should just do #449. I was thinking about #474 and it occurred to me that it is likely not a problem for most people since the tool dependencies.xml file should be used by default. That person likely installed the tool and explicitly opted for Conda instead of allowing the default dependency resolution. I'm not a -1 or anything though - up to y'all who maintain the tools.

nsoranzo:
Closing in favour of #449, I'm going to add my changes there.

PR #477, opened 2017-07-20
Be more explicit in description

caleb-easterly:
Specify that is greater than or equal to minimum and less than or equal to maximum. The ambiguity confused me while working in Galaxy.


bgruening:
Nice! I love these docs enhancements!

PR #227, opened 2015-09-18
Make bamtools_split output a dataset collection

pvanheus:
Replace the multiple output datasets of bamtools_split with a dataset collection output.

Also update the tests and the documentation to reflect this change.

PR #481, opened 2017-07-26
Move bamtools to legacy.

davebx:
Now that galaxyproject/tools-iuc#1359 has been merged, bamtools can be moved to legacy. This breaks #468 (which was included in 1359) and #227.


bgruening:
Thanks @davebx!

PR #485, opened 2017-09-20
Update to bwa 0.7.16 (which adds MateCigar for PE reads)

mvdbeek:


bgruening:
@mvdbeek thanks a lot!

PR #486, opened 2017-09-22
Update trinity_psc wrapper to more closely conform to IUC standards.

davebx:


nsoranzo:
@davebx Can't this be integrated with https://github.com/galaxyproject/tools-iuc/tree/master/tools/trinity ?

natefoo:
@nsoranzo The differences here are:

1. It makes use of a local filesystem for work. Thankfully, this version of the wrapper will only use it if the non-Galaxy-standard variable `$LOCAL` is set.
1. It calculates the memory settings in a way that is optimal for [Bridges](https://www.psc.edu/bridges), but may not apply for many systems, and also makes use of a non-Galaxy-standard variable `$CORE_MEMORY`
1. The version of Trinity we use is specific to Bridges (via the modules resolver) and includes minor modifications to run more efficiently on that system. This means that the newest version of Trinity may not always be available there even if it is in Conda.

That said, I would prefer that Main use the IUC tool and there are some possible solutions:

- 1 and 2: Define standard variables such as `$GALAXY_MEMORY`, `$GALAXY_MEMORY_PER_SLOT`, `$GALAXY_LOCAL_JOB_DIRECTORY` that can be set or manipulated via `<env>` tags in the job destination. I see that the IUC version of the tool uses `$TRINITY_MAX_MEMORY` for the memory limit purpose, but one-offing variables like this per tool is not ideal.
- 2: I'm not sure if the per-core calculation is the best default for a general purpose version of the wrapper, although it would allow for tools to dynamically set memory limits if `$GALAXY_MEMORY_PER_SLOT` is set, rather than having to explicitly set different amounts of memory allocated on each destination. In our case I could have the dynamic rule that handles our Trinity jobs set `$TRINITY_MAX_MEMORY` or whatever standard variable we create, though.
- 3: PSC has every major release since 2.0.6 installed and maintaining Trinity is a fairly high priority for them, so this is probably not a major concern.

natefoo:
Let's use `$TRINITY_SCRATCH_DIR` for now, and Galaxy admins can set it in the job conf as appropriate. In the future (galaxyproject/galaxy#4606) we can just use `$TMPDIR`.

galaxyproject/galaxy#4958 should allow `$GALAXY_MEMORY_MB` to take the place of `$TRINITY_MAX_MEMORY` with some changes, since it is an integer representing MB, and the amount used by Trinity should be slightly less, since it is a hard limit.

natefoo:
@davebx can you merge this with the IUC wrapper now that galaxyproject/galaxy#4958 is merged?

davebx:
@natefoo galaxyproject/tools-iuc#1620

I suppose this should now be deprecated in favor of IUC's trinity wrapper.

PR #494, opened 2017-11-17
added changeset_revision to samtools suite

bernt-matthias:
I guess that suites should define the changeset_revision of the contained tools. 

Advantages: 
- reproducibility: otherwise the same suite installed on different galaxies might give different results
- dependencies can be updated via the suite

### DIFF
--- a/suites/suite_samtools_1_2/repository_dependencies.xml
+++ b/suites/suite_samtools_1_2/repository_dependencies.xml
@@ -1,19 +1,19 @@
 <?xml version="1.0"?>
 <repositories description="A suite of Galaxy tools designed to work with version 1.2 of the SAMtools package.">
-    <repository name="data_manager_sam_fasta_index_builder" owner="devteam" />
-    <repository name="bam_to_sam" owner="devteam" />
-    <repository name="sam_to_bam" owner="devteam" />
-    <repository name="samtools_bam_to_cram" owner="devteam" />
-    <repository name="samtools_cram_to_bam" owner="devteam" />
-    <repository name="samtools_bedcov" owner="devteam" />
-    <repository name="samtools_calmd" owner="devteam" />
-    <repository name="samtools_flagstat" owner="devteam" />
-    <repository name="samtools_idxstats" owner="devteam" />
-    <repository name="samtools_mpileup" owner="devteam" />
-    <repository name="samtools_reheader" owner="devteam" />
-    <repository name="samtools_rmdup" owner="devteam" />
-    <repository name="samtools_slice_bam" owner="devteam" />
-    <repository name="samtools_sort" owner="devteam" />
-    <repository name="samtools_split" owner="devteam" />
-    <repository name="samtools_stats" owner="devteam" />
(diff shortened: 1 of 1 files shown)


mvdbeek:
I'm not sure this has any effect. I think the toolshed automatically sets the changeset_revision to the latest revision known when uploaded. You can see this in action by browsing https://toolshed.g2.bx.psu.edu/repository/browse_repository?id=42060580443e6eda#

mvdbeek:
> dependencies can be updated via the suite

That would be nice, but has never worked :(. On the upside the beta install UI will allow adding things to an installation queue, so it's not that painful anymore.

bgruening:
Imho we should move all samtools into one folder with a demultiplexing shed.yml and deprecate this suite file. Ideally moving all this into IUC.

nsoranzo:
Most samtools tools are already in tools-iuc: https://github.com/galaxyproject/tools-iuc/tree/master/tool_collections/samtools

bernt-matthias:
> I'm not sure this has any effect. I think the toolshed automatically sets the changeset_revision to the latest revision known when uploaded

Yes, and thereby different versions might be installed depending on the time. So different users might get different dependency versions. Thereby I suggest to set the revision explicitely. 

It seems that it has an effect when you look at https://github.com/galaxyproject/tools-devteam/blob/master/suites/suite_fastx_toolkit_0_0_13/repository_dependencies.xml implies fasta_clipping_histogram f2ab5b44870d (even if the TS only lists ec66ae4c269b .. which is strange in itself .. wondering what would happen when this is installed.)

Could someone point me to a place where I could read about shed.yml? I have not stumbled over this so far.

mvdbeek:
> Thereby I suggest to set the revision explicitely.

You can't, unless you change the toolshed code. I'd consider suites pretty much deprecated / an entrypoint to discover tools. We should work on updating more than one repository at the same time in a convenient way, which is the only thing suites can do.

> It seems that it has an effect when you look at https://github.com/galaxyproject/tools-devteam/blob/master/suites/suite_fastx_toolkit_0_0_13/repository_dependencies.xml implies fasta_clipping_histogram f2ab5b44870d (even if the TS only lists ec66ae4c269b .. which is strange in itself .. wondering what would happen when this is installed.)

This was set at the time of upload. If a changeset revision is not installable anymore (the revisions continue to exist!) the closest newer one will be used. 

> Could someone point me to a place where I could read about shed.yml? I have not stumbled over this so far.

These are only consumed by planemo, I don't think we have more documentation than http://galaxy-iuc-standards.readthedocs.io/en/latest/best_practices/shed_yml.html.

mvdbeek:
Actually I'm not so sure anymore, maybe what I described only applies to tool dependencies. I guess testing this wouldn't be a bad idea.

bernt-matthias:
Currently I have no galaxy to test this .. hopefully our test instance is up again next week and I remember to test.

mvdbeek:
For me the relevant test is to upload suites pointing to specific revisions to the toolshed. Galaxy will respect the indicated changeset revisions (except for the case in which they are not available, where Galaxy chooses the next available revision), of this I'm still pretty sure :).

Issue #111, opened 2015-04-20
vcftools dependencies: tabix and bgzip


lparsons:
It seems like the tools in the vcftools collection require both bgzip and tabix (see tools-devteam/tool_collections/vcftools/vcftools_subset/vcftools_subset.xml). However, they are not installed as a dependency from the toolshed. Are these tools that should be installed an the galaxy user's path (much like a believe samtools needs to be), or is this simply a dependency oversight?

bgruening:
@lparsons this is an oversight! Thanks for reporting!

nsoranzo:
bgzip and tabix are now included in [HTSlib](http://www.htslib.org/download/), so a dependency on (still-to-be-updated) samtools 1.2 would probably be more future-proof. See https://github.com/galaxyproject/tools-iuc/issues/146 .

aguynamedben:
+1 I ran into this following the CloudMan instructions. I failed to download my SSH key on cluster launch, now I have to restart my cluster in order to get a new key just so I can install bgzip.

PR #497, opened 2017-11-22
add tabix version

bgruening:


tool_collections/vcftools/vcftools_compare/vcftools_compare.xml
```diff
@@ -2,7 +2,7 @@
     <description>multiple VCF datasets</description>
 
     <requirements>
-        <requirement type="package">tabix</requirement>
+        <requirement type="package" version="0.2.5">tabix</requirement>
```
nsoranzo: Wouldn't it be better to use:
```
        <requirement type="package" version="1.6">htslib</requirement>
```
instead? 0.2.5 is pretty ancient and tabix has been distributed as part of htslib for ages.

bgruening: I just try to fix the old package. The multicontainer builds are failing because of this. Thats also the reason I have not bumped the tool version number.
Is this ok for you?

nsoranzo:
xref. https://github.com/galaxyproject/tools-devteam/issues/111

tool_collections/vcftools/vcftools_compare/vcftools_compare.xml
nsoranzo: That's not the only problematic requirement in this tool collection:
```
tool_collections/vcftools/vcftools_annotate/vcftools_annotate.xml
5:        <requirement type="binary">echo</requirement>

tool_collections/vcftools/vcftools_isec/vcftools_isec.xml
5:        <requirement type="package">tabix</requirement>

tool_collections/vcftools/vcftools_merge/vcftools_merge.xml
5:        <requirement type="package">tabix</requirement>

tool_collections/vcftools/vcftools_slice/vcftools_slice.xml
5:        <requirement type="binary">echo</requirement>

tool_collections/vcftools/vcftools_subset/vcftools_subset.xml
5:        <requirement type="package">tabix</requirement>
```

Also, `tool_collections/vcftools/vcftools_consensus/vcftools_consensus.xml` is already using htslib to provide tabix:
```
<requirement type="package" version="1.3.2">htslib</requirement>
```

Finally, I noticed that:
- only `vcftools_consensus` cites `10.1093/bioinformatics/btr330`
- documentation links in `<help>` could be updated to https://vcftools.github.io/examples.html .

The upstream vcftools seems to be maintained, so it may be worth to use a more sustainable solution?

PR #499, opened 2017-12-11
updated requirements and simplifications for xy_plot

bernt-matthias:
R does not exist in conda, r-base does

Tool now uses Rscript instead of shell script.


nsoranzo:
@bernt-matthias I've pushed a commit to your branch to remove the `fontconfig` requirement, which I think is unnecessary, and update to the latest `r-base` available in Conda. I hope that's OK with you.

bernt-matthias:
happens also with 17.09

actually the call to conda gives the error: 

 `/home/berntm/miniconda3/bin/conda create -y --override-channels --channel iuc --channel bioconda --channel conda-forge --channel defaults --channel r --name __r-base@3.4.2 r-base=3.4.2`

```
Fetching package metadata ...............
Solving package specifications: .


PackageNotFoundError: Package not found: '' Dependencies missing in current linux-64 channels: 
  - r-base 3.4.2* -> libgcc-ng 7.2.0.*
  - r-base 3.4.2* -> libgfortran-ng 7.2.0.*
  - r-base 3.4.2* -> libstdcxx-ng 7.2.0.*
  - r-base 3.4.2* -> ncurses 6.0.*

Close matches found; did you mean one of these?

    libgcc-ng: libgcc
    libgfortran-ng: libgfortran

You can search for packages on anaconda.org with

    anaconda search -t conda ncurses

(and similarly for the other packages)

You may need to install the anaconda-client command line client with

    conda install anaconda-client
```

PR #500, opened 2017-12-18
bwa: added a regular expression to stdio

bernt-matthias:
This happens when unexpected input is provided. For instance a fa.tar file is used (I also read that spaces in the sequence may cause this message).


mvdbeek:
This is now maintained at https://github.com/galaxyproject/tools-iuc/tree/master/tools/bwa, can you open a PR there ?

bernt-matthias:
Sure. Got confused, because devteam is still the owner in the TS.

PR #502, opened 2018-02-13
remove lastz

bgruening:


nsoranzo:
I think we usually move them to `legacy/`

bernt-matthias:
Hi @bgruening  what was the reason for this PR? I just stumbled over lastz because it fails for the CI that just ran for the tool: https://github.com/galaxyproject/tools-devteam/pull/558

Should we just (re)move it or fix it first?

nsoranzo:
It's already in tools-iuc.

bernt-matthias:
I see. Should we just move it to legacy (we should also add `legacy` to the exclude list in CI or rename the folder to deprecated for consistency with IUC repo structure)?

nsoranzo:
@bernt-matthias We are not putting in legacy only stuff that is not worth to migrate to tools-iuc, see https://github.com/galaxyproject/tools-devteam/pull/532 .

bernt-matthias:
Thanks for the clarification.

PR #503, opened 2018-02-19
updated velvet wrappers

bernt-matthias:
Some first steps toward an up-to-date velvet wrapper

- use GALAXY_SLOTS
- removed python wrapper script
- allow input of separate paired end reads
- allow larger k-mer size up to the limit of the conda requirement (note: even inputs are handled by velvet by subtracting 1 -- should be fine)

Open: 
- testing 
- but seemingly the test data was lost somewhere. Any idea where this could be? 
- the parameters of the columbus extensions could be implemented (https://www.ebi.ac.uk/~zerbino/velvet/Columbus_manual.pdf)

I'm unsure how the output directory is handled. Seemed to work before. I just do not understand it. Could someone point me to the documentation?

Any other ideas. I will take care of velvetg later.


tools/velvet/velveth.xml
```diff
@@ -1,70 +1,97 @@
...
-           '$out_file1' '$out_file1.extra_files_path'
-           $hash_length
-           $strand_specific
-           #for $i in $inputs
-                ${i.file_format}
-                ${i.read_type}
-                ${i.input}
-           #end for
-  </command>
+  <version_command>velveth 2&gt;&amp;1 | grep "Version" | sed -e 's/Version //'</version_command>
```
bgruening: You could also use CDATA here.

tools/velvet/velveth.xml
```diff
@@ -1,70 +1,97 @@
...
+    export OMP_NUM_THREADS="\${GALAXY_SLOTS:-1}" &&
+    export OMP_THREAD_LIMIT="\${GALAXY_SLOTS:-1}" &&
+    velveth
+      '$out_file1.extra_files_path'
+      $hash_length
+      #for $i in $inputs
+          ${i.file_format}
+          ${i.read_type}
+          #if str( $i.input_type.input_type_selector ) == "paireds":
+              -separate $i.input1 $i.input2
```
bgruening: single quotes for all data and text params.

tools/velvet/velveth.xml
```diff
@@ -1,70 +1,97 @@
...
+          #else if str( $i.input_type.input_type_selector ) == "pairedi":
+              -interleaved ${i.input}
+          #else:
+              ${i.input}
```
bgruening: single quotes

tools/velvet/velveth.xml
```diff
@@ -1,70 +1,97 @@
...
-      <option value="15">15</option>
-      <option value="17">17</option>
-      <option value="19">19</option>
-      <option value="21" selected="yes">21</option>
-      <option value="23">23</option>
-      <option value="25">25</option>
-      <option value="27">27</option>
-      <option value="29">29</option>
-    </param>
+      <param label="Hash Length" name="hash_length" type="integer" value="21" min="11" max="191" help="k-mer length in base pairs of the words being hashed."/>
```
bgruening: is it still true that we need to compile in the max kmer length?

bernt-matthias: seems so. at least velvet outputs this on startup. was wondering if we can parse this and use the value in the form. is there such a possibility?

bernt-matthias:
I would definitely like to add a test. Since there is already a test (but the data is missing) I was wondering if the data might still be somewhere. Any ideas? Otherwise I would take these: http://galaxyproject.github.io/training-material/topics/assembly/tutorials/general-introduction/tutorial.html

bernt-matthias:
There is the possibility to input raw (and raw.gz)  data. Any idea if there is a Galaxy data type for this?

bernt-matthias:
For raw: I can only guess from a peek into the velvet sources that it is sequence data without any header info. I guess this is used only rareley. Also for raw no separate input of paired data seems possible. I guess we leave it out..?

nsoranzo:
@bernt-matthias I've found the test-data: http://lists.bx.psu.edu/pipermail/galaxy-dev/attachments/20100603/f9557958/attachment-0001.zip , see https://lists.galaxyproject.org/pipermail/galaxy-dev/2010-June/002760.html

tools/velvet/velveth.xml
Slugger70: I always just compile velvet with Max kmer size 191, omp turned on and categories 4 as this cover most things... No point using the defaults, they're useless.. they were conceived a long time ago

Slugger70:
I'd leave Raw out. I've never seen it used.. Bioconda Velvet is compiled with MAXKMERLENGTH=191, CATEGORIES=4, OPENMP=1, LONGSEQUENCES=1. You use the OPENMP settings by setting a couple of env vars. See the VelvetOptimiser wrapper for details. 
``` <command detect_errors="exit_code"><![CDATA[
        export OMP_NUM_THREADS=2 &&
        export OMP_THREAD_LIMIT=2 &&
        VelvetOptimiser.pl
            -t "\${GALAXY_SLOTS:-1}"
            -s $start_kmer
            -e $end_kmer
            -d out
            -f "
```

I couldn't figure out how to make the omp_num_threads set in a more dynamic way..

bernt-matthias:
The velveth wrapper should now be functional. 
- threading is enabled with `export OMP_NUM_THREADS="\${GALAXY_SLOTS:-1}" &&`. Unfortunately velvet uses one more thread than specified this way, but setting `OMP_THREAD_LIMIT` the same way seems to be dysfunctional in if the value is 1. **Can we compute here? Then we could +1?**
- I took a down sampled version of the GTN test data

Some questions on velvetg
- There was `<option value="yes" selected="${input.metadata.paired_end_reads}">Yes</option>` Does such a thing work? Linting fails with this. 
- The test currently fails because the `ftype="velvet"` seems not to work. The data type seems odd to me: it consists of a (quite useless) html file and the Sequences and Roadmaps file in a sub directory. The velveth wrapper currently relies on the fact that the html file is seemingly autogenerated by the velvet data type class? (In the original version of this wrapper the velveth_wrapper.py script did this)
- I guess some of the conditions are unnecessary and can be replaced by simple bools?

bgruening:
@bernt-matthias I tried to enable testing for this repo.

nsoranzo:
@bernt-matthias You can compute using bash syntax, see e.g. https://github.com/galaxyproject/tools-iuc/blob/master/tools/trinity/trinity.xml#L11

lparsons:
Could/should this be moved to https://github.com/galaxyproject/tools-iuc?

bernt-matthias:
Could: Yes. Should: I don't know. 

Any comments on the question in my last comment? Then I could finish the work in this PR.

bgruening:
Moving to IUC: yes. But I think we can also do this after this PR.
I can not answer your questions, but @Slugger70 probably can. I'm wondering if velvet is worth to update nowadays.

bernt-matthias:
I think its worth -- and if its only for the GTN tutorials.

bernt-matthias:
Travis still stumbles over fatal_oom ...?

nsoranzo:
@bernt-matthias Should be fixed by https://github.com/galaxyproject/planemo/pull/866 and a new Planemo release.

bernt-matthias:
the linting seems to work now for iuc, maybe we can also restart the checks here?

bgruening:
Restarted!

bernt-matthias:
I hope that I fixed the last bugs. Odd thing is that linting now fails locally for me ..

bgruening:
ping @Slugger70 :)

tools/velvet/velveth.xml
bernt-matthias: done

tools/velvet/velveth.xml
bernt-matthias: done

tools/velvet/velveth.xml
bernt-matthias: done

bernt-matthias:
closed via https://github.com/galaxyproject/tools-devteam/pull/564

Issue #340, opened 2016-02-24
cummeRbund 1.0.1 error


methylome:
I have no clue why the following errors occurred.
- When loading:
galaxy.tools.parameters.basic DEBUG 2016-02-24 04:04:17,191 Error determining dynamic options for parameter 'gene_id' in tool 'toolshed.g2.bx.psu.edu/repos/devteam/cummerbund/cummeRbund/1.0.1':
Traceback (most recent call last):
  File "/home/galaxy/dev/lib/galaxy/tools/parameters/basic.py", line 879, in get_options
    return eval( self.dynamic_options, self.tool.code_namespace, call_other_values )
  File "<string>", line 1, in <module>
NameError: name 'input_database' is not defined

- When executing:
galaxy.tools.deps DEBUG 2016-02-24 04:04:59,270 Building dependency shell command for dependency 'CUMMERBUND_SCRIPT_PATH'
galaxy.tools.deps DEBUG 2016-02-24 04:04:59,270 Find dependency CUMMERBUND_SCRIPT_PATH version None
galaxy.tools.deps DEBUG 2016-02-24 04:04:59,299 Resolver tool_shed_packages returned <galaxy.tools.deps.resolvers.galaxy_packages.GalaxyPackageDependency object at 0x7f173c3de990> (isnull? False)
galaxy.tools.deps DEBUG 2016-02-24 04:04:59,300 Building dependency shell command for dependency 'R'
galaxy.tools.deps DEBUG 2016-02-24 04:04:59,300 Find dependency R version 3.1.2
galaxy.tools.deps DEBUG 2016-02-24 04:04:59,301 Resolver tool_shed_packages returned <galaxy.tools.deps.resolvers.galaxy_packages.GalaxyPackageDependency object at 0x7f173c3de090> (isnull? False)
galaxy.tools.deps DEBUG 2016-02-24 04:04:59,301 Building dependency shell command for dependency 'cummeRbund'
galaxy.tools.deps DEBUG 2016-02-24 04:04:59,301 Find dependency cummeRbund version 2.8.2
galaxy.tools.deps DEBUG 2016-02-24 04:04:59,301 Resolver tool_shed_packages returned <galaxy.tools.deps.resolvers.galaxy_packages.GalaxyPackageDependency object at 0x7f173c3dee50> (isnull? False)
galaxy.jobs.runners ERROR 2016-02-24 04:04:59,515 (137) Failure preparing job
Traceback (most recent call last):
[...]

PR #506, opened 2018-06-08
Update cummeRbund to retrieve genes and samples from input metadata.

davebx:
This depends on galaxyproject/galaxy#6302 and should resolve galaxyproject/usegalaxy-playbook#110 and #340


bernt-matthias:
Is this still relevant? If so: could you rebase/merge? Then I could have a look.

PR #507, opened 2018-06-08
this tool seems to be broken

bgruening:
I wonder if it ever worked.


davebx:
@bgruening it would have worked the other way, but this brings it up to better standards. In fact the ./ in front of `blast` is not technically necessary, since `ln` with no path for the target will put it in the CWD.

bgruening:
@davebx thanks Dave. I think the problem was a space in the blast db name or something like that.

davebx:
@bgruening that makes perfect sense, and illustrates why we decided to make quoted paths a standard. Thanks for the fix.

PR #508, opened 2018-06-25
more precice description of the concat tool

bernt-matthias:


tool_collections/gops/concat/concat.xml
```diff
@@ -1,5 +1,5 @@
-<tool id="gops_concat_1" name="Concatenate" version="1.0.1">
-    <description>two datasets into one dataset</description>
+<tool id="gops_concat_1" name="Concatenate" version="1.0.2">
```
bgruening: I guess in this case we do not need a bump.

PR #511, opened 2018-08-21
WIP: Correlation work

bgruening:
needs to be ported to rpy2


tools/correlation/cor.xml
```diff
@@ -1,33 +1,48 @@
...
+        '$input1'
+        '$out_file1'
+        $numeric_columns
+        $method
+    </command>
+    <inputs>
+        <param format="tabular" name="input1" type="data" label="Dataset" help="Dataset missing? See TIP below"/>
+        <param name="numeric_columns" label="Numerical columns" type="text" multiple="True"
+            data_ref="input1" help="Multi-select list - hold the appropriate key while clicking to select multiple columns" />
+        <!--param name="numeric_columns" label="Numerical columns" type="data_column" numerical="True" multiple="True"
```
bgruening: @guerler do you remember that we talk once about the data_column parameter and that it would be nice to allow an `overwrite mode` like in the workflow form. So a user could choose to insert numbers instead of choosing the fields. 
I get this request now a lot in communities that do more statistics on large files (>30 columns).

This PR demonstrates a simple tool that can take `2-3` as an indicator of the columns instead of `2,3`. Tests are passing, maybe you can use this to explore some UI options? The idea is to offer the user some more advances selections like `2,5,10-30,40` in an advanced mode next to the traditional `data_column` view.

Issue #390, opened 2016-09-14
Update SICER to include more genome builds (ideally all on the instance)


jennaj:
The list of builds is outdated. At a minimum add the latest builds from the same organisms already included: mm10, hg38, dm6, sacCer3, rn5, rn6. 

The "pombe" and "tair8" are not dbkeys in current Galaxy releases ... (??) but there are alternatives.

Quote from sicer_wrapper.py

> #HACK! FIXME: allow using all specified builds, would currently require hacking SICER's "GenomeData.py" on the fly.
> VALID_BUILDS = [ 'mm8', 'mm9', 'hg18', 'hg19', 'dm2', 'dm3', 'sacCer1', 'pombe', 'rn4', 'tair8' ]

ACastanza:
I'd also like to see this updated, seems a little weird that so many of these tools (and related publications) are still using mm9.

jennaj:
ping @martenson @davebx @nekrut 

Are there plans to update this tool? Was asked about again here: https://biostar.usegalaxy.org/p/28442/. Thanks!

PR #514, opened 2018-09-19
Add mm10 to sicer

ccedmendoza:
Hi,
I had to modify the sicer_wrapper files as well as GenomeData.py so I could use it with mm10.
I thought I should share it but I'm not sure how to add the GenomeData.py file modifications

### DIFF
--- a/tools/sicer/GenomeData.py
+++ b/tools/sicer/GenomeData.py
@@ -1,0 +1,206 @@
+#!/usr/bin/env python
+# Copyright (c) 2007 NHLBI, NIH
+# Copyright (c) 2010 The George Washington University
+# Authors: Chongzhi Zang, Weiqun Peng, Dustin E Schones and Keji Zhao
+#
+# This software is distributable under the terms of the GNU General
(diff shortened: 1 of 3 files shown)


ccedmendoza:
This has to do with issue #390

nekrut:
@blankenberg?

blankenberg:
The biggest issue here has to do with the fact that sicer (`GenomeData.py`) itself needs to be modified to work with each new genome build. Apparently I wrote to the sicer authors  around 2011 (see e.g. [here](https://lists.galaxyproject.org/pipermail/galaxy-dev/2012-March/008889.html)) to see about them updating sicer to accept `.len` files instead of having hard coded lengths. I'm assuming this hasn't been done.

If we are going the modification route, I would suggest modifying the underlying sicer code to accept `.len` files instead of just adding more hard coded values. This should be really quite easy to do.

If so, it is probably worth creating a separate github repo to track the changes and making conda recipes, etc. Would need to double check the licensing of course as well.

Unfortunately, I would not suggest merging this PR as-is.

PR #515, opened 2018-09-24
Fix output format for coverage and join

nsoranzo:
I broke these in commit cae3e05d02e60f595bb8b6d77a84f030e9bd1689 .

### DIFF
--- a/tool_collections/gops/join/join.xml
+++ b/tool_collections/gops/join/join.xml
@@ -27,7 +27,7 @@ python '$__tool_directory__/gops_join.py'
         </param>
     </inputs>
     <outputs>
-        <data name="output" format_source="input1" metadata_source="input1" />
+        <data name="output" format="interval" metadata_source="input1" />
     </outputs>
     <tests>
         <test>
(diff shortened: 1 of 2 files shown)


tool_collections/gops/join/join.xml
```diff
@@ -27,7 +27,7 @@ python '$__tool_directory__/gops_join.py'
         </param>
     </inputs>
     <outputs>
-        <data name="output" format_source="input1" metadata_source="input1" />
+        <data name="output" format="interval" metadata_source="input1" />
```
mvdbeek: Isn't this fix making it less flexible ?
E.g bed input1 and bed input2 becoming interval ?

nsoranzo: No, the joined output is not in `bed` format, it has 12 columns, 6 columns from the first input and 6 from the second, so columns 7-12 are not what is described in e.g https://genome.ucsc.edu/FAQ/FAQformat.html#format1

In fact, if you upload the output test data to Galaxy, they are all sniffed as `interval`, not `bed`.

Similarly for the coverage tool.

PR #516, opened 2018-10-05
Column maker header

lecorguille:
It's the same idea as for sort and join: the possibility to conserve the header of the table file.
The code may be better, I guess. So don't hesitate to give me advice.


lecorguille:
Hum ... I will add a param to fill the new column name. It will be better than the `$cond` c4/c5

lecorguille:
Done

tools/column_maker/column_maker.xml
```diff
@@ -1,23 +1,47 @@
...
         python $__tool_directory__/column_maker.py
-            $input $out_file1
+            data column_maker_output
             "$cond"
             $round
             ${input.metadata.columns}
-            "${input.metadata.column_types}"
-    </command>
+            "${input.metadata.column_types}" &&
+        cat header column_maker_output > $out_file1
```
bgruening: outfile needs to be single-quoted

tools/column_maker/column_maker.xml
```diff
@@ -1,23 +1,47 @@
...
         </param>
+        <conditional name="header_lines_conditional">
+            <param name="header_lines_select" type="select" label="Skip a header line" help="# characters are already considered as comments and kept" >
+                <option value="no" >no</option>
+                <option value="yes" >yes</option>
+            </param>
+            <when value="no">
+            </when>
+            <when value="yes">
+                <param name="header_new_colonne_name" type="text" value="New Column" label="The new column name" />
```
bgruening: should that be `header_new_column_name`?

bgruening:
Thanks @lecorguille!

PR #517, opened 2018-10-08
Rename .tt_blacklist to .tt_skip (tools we don't test on Travis)

peterjc:
The terms 'whitelist' and 'blacklist' for inclusion or exclusion have undesirable associations, making it better to use neutral language.

See also "Be careful in the words that we choose" in the main Galaxy Project Code of Conduct,
https://github.com/galaxyproject/galaxy/blob/dev/CODE_OF_CONDUCT.md

This is initially the same changes as proposed for ``tools-iuc`` in https://github.com/galaxyproject/tools-iuc/pull/2123 - assuming that is accepted, I'll update this pull request to match (e.g. alternative filename).


peterjc:
Ah, ``planemo ci_find_repos`` needs updating too - I'll fix that.

PR #518, opened 2018-10-22
Condafy sicer

ccedmendoza:
SICER needs scipy to run so I added a requirement to the XML wrapper.


tools/sicer/sicer_wrapper.xml
```diff
@@ -28,6 +28,7 @@
   </command>
   <requirements>
     <requirement type="package" version="1.1">SICER</requirement>
+    <requirement type="package" version="1.1.0=py27hd20e5f9_0">scipy</requirement>
```
bgruening: just 1.1.0

tools/sicer/sicer_wrapper.xml
```diff
@@ -28,6 +28,7 @@
   </command>
   <requirements>
     <requirement type="package" version="1.1">SICER</requirement>
+    <requirement type="package" version="1.1.0">scipy</requirement>
```
nsoranzo: That's actually not needed since the sicer bioconda package already depends on scipy: https://github.com/bioconda/bioconda-recipes/blob/master/recipes/sicer/meta.yaml#L27

ccedmendoza: I think it's because it never uses the SICER bioconda package. It installs the toolshed package and the bioconda package but it doesn't activate the SICER conda environment. So the scipy is missing in a clean Galaxy instance.

nsoranzo: That's because the `tool_dependencies.xml` file was present in the ToolShed repository, once removed this will work fine.

nsoranzo:
Can you also please remove `tools/sicer` from ` .tt_skip`, so it gets tested?

mvdbeek:
thanks @ccedmendoza!

PR #523, opened 2019-06-24
Column maker improvements

mvdbeek:
Modernize the script, use python 3.7 and coreutils (for `tr`) as dependency and allow single column tabular input, fail with non-zero exit code in `stop_err`. I want to include this tool in https://github.com/galaxyproject/training-material/pull/1393


nsoranzo:
Should we move this too tools-iuc too?

mvdbeek:
Yeah, let's do that. I think this is a widely used tool.

mvdbeek:
Alright, done in https://github.com/galaxyproject/tools-iuc/pull/2469

PR #524, opened 2019-09-10
Update compute lengths tool to support data tables

erasche:
Wanted to allow users to pull out lengths of genomes for the Circos tool, since currently there's no way for users to see this table, and the users need to select which chromosomes they want to visualise.

For the duplicated options - I tried embedding and having an alongside macros.xml and in both cases it killed my tool watching, and would not load complaining about invalid XML in planemo (which passes xmllint fine) and so I've just duplicated them.


tools/fasta_compute_length/fasta_compute_length.xml
```diff
@@ -1,42 +1,74 @@
...
     <inputs>
-        <param name="input" type="data" format="fasta" label="Compute length for these sequences"/>
-        <param name="keep_first" type="integer" value="0" label="How many title characters to keep?" help="'0' = keep the whole thing"/>
-        <param name="keep_first_word" type="boolean" truevalue="id_only" falsevalue="id_and_desc"
-            selected="false" label="Strip fasta description from header?"
-            help="Stripping the description will truncate the fasta header to just the sequence ID. Otherwise the header description will be kept. This step is done before the 'How many characters to keep' option."/>
+        <conditional name="ref">
+            <param name="ref_source" type="select" label="Sequences">
+                <option value="history" selected="True">From History</option>
+                <option value="dbkey">Locally Cached (pre-built length files)</option>
```
bgruening: It is, unfortunately, not that easy. You need to add some boilerplate like tool_data_table_conf files for sample and test data... https://github.com/galaxyproject/tools-iuc/tree/61e1699e65d6fd9f4f73650ed8463b37cd701344/tools/idba_ud

bgruening: And a tool-data folder: https://github.com/galaxyproject/tools-iuc/tree/61e1699e65d6fd9f4f73650ed8463b37cd701344/tools/idba_ud/tool-data

erasche: Done.

PR #525, opened 2019-09-11
velveth_xml more hash size

davidchristiany:
Very simple modification of the velveth wrapper in order to add more choices in hash length.


tools/velvet/velveth.xml
```diff
@@ -29,6 +29,66 @@
...
+      <option value="131">131</option>
+      <option value="133">133</option>
+      <option value="135">135</option>
+      <option value="137">137</option>
+      <option value="139">139</option>
+      <option value="141">141</option>
+      <option value="143">143</option>
+      <option value="145">145</option>
+      <option value="147">147</option>
+      <option value="149">149</option>
```
bgruening: You can actually go until 191 https://github.com/bioconda/bioconda-recipes/blob/master/recipes/velvet/build.sh#L3 
if you depend on the bioconda package. This is a compile-time option.

PR #527, opened 2019-09-30
Actually use some gffread parameters

erasche:
@abretaud pointed me in the way of gffread and in trying to use it, noticed that (at least) these three parameters were not included in the code block (despite being quite well documented in inputs + help section.)

edit: title doesn't sound like how I want, sorry, please read it with good intentions/feelings.


erasche:
Ready for review if someone has a minute :)

PR #532, opened 2019-10-29
Migrate gffread to IUC

lparsons:
Migrate the gffread utility to the IUC via https://github.com/galaxyproject/tools-iuc/pull/2665


bgruening:
Mh, I considered legacy really as lagacy-tools that are not maintained. I don't see a point in moving them. :(

martenson:
`/legacy` is exactly for that purpose, those are tools that made it to the IUC already.

bgruening:
@martenson merge or shift them to legacy? Your call.

nsoranzo:
I think it was agreed to remove migrated tools from now on and clean up `legacy/` .

Issue #536, opened 2019-12-06
Fasta filer by length fails when max_length !=0


Takadonet:
Issue reported by end user that they could not filter out large contigs from their fasta.  They had a contig file with 2 contigs over 2Mbp and remaining under 50kbp . When values were set to the following:

![contigs](https://user-images.githubusercontent.com/64732/70344099-7d63cf00-181e-11ea-81b8-2a8d6d0d38ce.png)

All contigs were filtered **in**.

Looks like the issue is that we have a[ 'one off' error](https://github.com/phac-nml/tools-devteam/blob/master/tools/fasta_filter_by_length/fasta_filter_by_length.py#L41) where tmp_size will stop counting when over max_length but not when it's equal. Hence if your maximum length is evenly divisible by the line length of your fasta record, it will always filter in!

bernt-matthias:
Seems fixed.

PR #537, opened 2019-12-06
Bug fix: Max length evenly divisible by line length

Takadonet:
Address #536


tools/fasta_filter_by_length/fasta_filter_by_length.xml
```diff
@@ -22,6 +22,12 @@
 			<param name="max_length" value="0" />
 			<output name="output" file="fasta_tool_filter_length_1.out" />
 		</test>
+		<test>
+			<param name="input" value="4.fasta" />
+			<param name="min_length" value="0" />
+			<param name="max_length" value="60" />
```
nsoranzo: Shouldn't this be more stringent, i.e.:
```suggestion
			<param name="max_length" value="54" />
```

Takadonet: Deleted my previous comment... It would be more stringent but issue would not cause the test to fail.  When the maximum length is evenly divided by the line length, issue will cause larger contigs to be filtered in. 

To use value 54, we would have to change DNA line length to 54.

Takadonet:
@nsoranzo need anything else from me for this to be merged?

nsoranzo:
Thanks @Takadonet !

Takadonet:
no thank you! @nsoranzo

PR #545, opened 2020-02-05
Add requirements to sam_pileup.py, fix script

mvdbeek:


tools/sam_pileup/sam_pileup.py
```diff
@@ -21,50 +21,55 @@
...
+import tempfile
 from bx.cookbook import doc_optparse
 
-def stop_err( msg ):
-    sys.stderr.write( '%s\n' % msg )
-    sys.exit()
+
+def stop_err(msg, e=None):
+    if e is not None:
+        raise e
```
nsoranzo: This way `msg` is never printed.

tools/sam_pileup/.shed.yml
```diff
@@ -9,5 +9,6 @@ long_description: |
   dataset with consensus is generated.
 name: sam_pileup
 owner: devteam
+homepage_url: https://github.com/galaxyproject/tools-devteam/tree/master/tools/sam_pileup
```
nsoranzo:
```suggestion
homepage_url: https://www.htslib.org/
```

nsoranzo:
After spending sometime cleaning up the overlong Python code, I realised this doesn't need to use a Python script at all... I'll work on that later or tomorrow.

mvdbeek:
It's probably not worth investing too much time though, noone should be using pileup format in 2020 anyway.

nsoranzo:
@mvdbeek I got rid of the Python, can you double check the changes?

mvdbeek:
Looks good, thanks @nsoranzo !

PR #549, opened 2020-02-27
Update CI for 20.01 (and other tools-iuc changes)

mvdbeek:


.travis.yml
```diff
@@ -13,6 +13,7 @@ jobs:
       addons: {}
       before_install: skip
       install:
+        - pip install planemo
```
mvdbeek: Guess I was a bit radical there 🤣

PR #552, opened 2020-03-04
Convert bytes to string for for stderr output in unified_genotyper.py

moskalenko:
A naive fix for

  File "toolshed.g2.bx.psu.edu/repos/devteam/unified_genotyper/66dd4d4c1743/unified_genotyper/gatk_wrapper.py", line 126, in <module>

    if __name__=="__main__": __main__()

  File "toolshed.g2.bx.psu.edu/repos/devteam/unified_genotyper/66dd4d4c1743/unified_genotyper/gatk_wrapper.py", line 115, in __main__

    stderr_target.write( chunk )

TypeError: write() argument must be str, not bytes


bgruening:
@moskalenko can you add a test for this case? Is this possible?

PR #558, opened 2020-05-16
introduce github actions for PR and CI

bernt-matthias:
Latest state from IUC.


.github/workflows/ci.yaml
```diff
@@ -0,0 +1,136 @@
...
+on:
+  schedule:
+    # Run at midnight every monday
+    - cron: '0 0 * * 1'
+env:
+  GALAXY_RELEASE: release_20.05
+jobs:
+  setup:
+    name: Setup cache
+    if: github.repository == 'galaxyproject/tools-iuc'
```
nsoranzo, line 11:
```suggestion
    if: github.repository == 'galaxyproject/tools-devteam'
```

bernt-matthias: Thanks @nsoranzo ..

Issue #546, opened 2020-02-05
GATK 1.4 tools need to be run in a Py2 environment (all)


jennaj:
MTS: https://toolshed.g2.bx.psu.edu/view/devteam/suite_gatk_1_4/3e6a4fcbf335

ping @mvdbeek @natefoo

bernt-matthias:
Also there seems to be no conda package for version 1.4. (which is horribly outdated?) ..

nekrut:
we've been talking about getting rid of it for a long time ... there is
absolutely no reason to maintain this

Anton Nekrutenko
Professor of Biochemistry
and Molecular Biology
Penn State University
*http://nekrut.bx.psu.edu <http://nekrut.bx.psu.edu>*
(814) 826-9628

PR #563, opened 2020-06-05
Move old gatk tools to legacy/

nsoranzo:
Also:
- Move tools in legacy to a `tools/` subdir
- Update and sort `.tt_skip` file

Close https://github.com/galaxyproject/tools-devteam/issues/546 .


bernt-matthias:
One could also rename `legacy` to `deprecated ` (in analogy to IUC)?

nsoranzo:
> One could also rename `legacy` to `deprecated ` (in analogy to IUC)?

I don't mind either way, but all other changes here are still needed.

nsoranzo:
> Then I would prefer `deprecated` in order to keep it as close to IUC as possible.

OK, done in a rebase.

bernt-matthias:
Do you think anyone else cares about this? Otherwise I would merge.

nsoranzo:
> Do you think anyone else cares about this? Otherwise I would merge.

I think you can safely go ahead if you are happy with it (we can always revert).

PR #569, opened 2020-07-15
Some cosmetic changes to mergeCols

mvdbeek:
I set out to fix what @yvanlebras reported on gitter:
```
There is an issue regarding the "Merge Columns" tool https://ecology.usegalaxy.eu/root?tool_id=toolshed.g2.bx.psu.edu/repos/devteam/merge_cols/mergeCols1/1.0.2 . Using 1.0.2 version, all the table is transformed as a one line file. Using 1.0.1 version https://ecology.usegalaxy.eu/root?tool_id=toolshed.g2.bx.psu.edu/repos/devteam/merge_cols/mergeCols1/1.0.1 is ok
history where you can look at this issue (dataset 14 with version 1.0.2, dataset 15 with 1.0.1): https://ecology.usegalaxy.eu/u/ylebras/h/test-training-session
```

I thought the issue was with `print(line, file=outfile)` not adding newlines,
but `\n` is actually the default end character. I can't reproduce Yvan's issue,
but might be a good idea to use `outfile.write()` with explicit newlines, soooo
... maybe let's try this, unless someone can spot what is wrong with Yvan's
example.

### DIFF
--- a/.tt_skip
+++ b/.tt_skip
@@ -56,7 +56,6 @@ tools/maf_cpg_filter
 tools/mapping_to_ucsc
 tools/megablast_wrapper
 tools/megablast_xml_parser
-tools/merge_cols
 tools/microsatellite_birthdeath
 tools/microsats_alignment_level
 tools/microsats_mutability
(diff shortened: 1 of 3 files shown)


nsoranzo:
Looks like that this tool is not tested, can you remove `tools/merge_cols` from `.tt_skip` please?

PR #570, opened 2020-08-25
py3k megablast

bgruening:
If that passes can I request a review @peterjc :)

### DIFF
--- a/tools/megablast_xml_parser/megablast_xml_parser.py
+++ b/tools/megablast_xml_parser/megablast_xml_parser.py
@@ -1,16 +1,12 @@
 #!/usr/bin/env python
-    
-import sys, os, re
 
-if sys.version_info[:2] >= ( 2, 5 ):
-    import xml.etree.cElementTree as ElementTree
-else:
-    from galaxy import eggs
-    import pkg_resources; pkg_resources.require( "elementtree" )
-    from elementtree import ElementTree
+import re
+import sys
 
-def stop_err( msg ):
-    sys.stderr.write( "%s\n" % msg )
+import xml.etree.cElementTree as ElementTree
+
+def stop_err(msg):
+    sys.stderr.write("%s\n" % msg)
     sys.exit()
 
 def __main__():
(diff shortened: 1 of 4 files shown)


tools/megablast_xml_parser/megablast_xml_parser.py
```diff
@@ -1,78 +1,71 @@
...
+import re
+import sys
+import xml.etree.cElementTree as ElementTree
 
-def stop_err( msg ):
-    sys.stderr.write( "%s\n" % msg )
+
+def stop_err(msg):
+    sys.stderr.write("%s\n" % msg)
     sys.exit()
```
bernt-matthias, line 10: Just `sys.exit("%s\n" % msg") `

tools/megablast_xml_parser/megablast_xml_parser.py
```diff
@@ -1,78 +1,71 @@
...
+        stop_err("Invalid data format.")
     # turn it into an iterator
-    context = iter( context )
+    context = iter(context)
     # get the root element
     try:
-        event, root = context.next()
-    except:
-        stop_err( "Invalid data format." )
+        event, root = context.__next__()
```
bernt-matthias, line 38: Is this really a `__` function?

peterjc: Why not ``event, root = next(context)`` which is Python 3 style?

tools/megablast_xml_parser/megablast_xml_parser.py
peterjc: I think ``sys.exit(msg)`` should work just as well:

```
$ python -c "import sys; sys.exit('Bye')"
Bye
```

nsoranzo: > I think `sys.exit(msg)` should work just as well:

Yes, and there's no need to define a `stop_err` function to wrap that.

bernt-matthias (approved):

bernt-matthias:
Anything open here? Or reasons against merging?

Issue #540, opened 2020-01-31
peakcalling_macs needs python 2 dependency


mvdbeek:

bernt-matthias:
Even worse: it needs a macs dependency :) .. which is not available in bioconda. Since macs2 is in IUC I'm suggesting to deprecate the tool.

PR #573, opened 2020-09-09
deprecate peakcalling macs

bernt-matthias:
macs is not in bioconda. macs2 is in IUC

fixes https://github.com/galaxyproject/tools-devteam/issues/540


mvdbeek:
I think the coursera course still uses macs ... but it really shouldn't. I don't know if we'll ever update the course, so 👍

PR #574, opened 2020-09-16
deprecate broken samtools pileup

bernt-matthias:
samtools mpileup is now at IUC

seems to be the last tool in the devteam repo that fails the weekly tests.


bgruening:
ping @wm75

nsoranzo:
pileup is different from mpileup (both as a tool and as a format), but the former has been deprecated since samtools 0.1.17 (6 July, 2011), so I think it's still fine to deprecate it.

Can you please move the `sam_pileup` directory in the `tools/` subdirectory of `deprecated/` ?

bernt-matthias:
> Can you please move the `sam_pileup` directory in the `tools/` subdirectory of `deprecated/` ?

Yep. Also moved sam_merge -- which I misplaced as well :(

nsoranzo:
> > Can you please move the `sam_pileup` directory in the `tools/` subdirectory of `deprecated/` ?
> 
> Yep. Also moved sam_merge -- which I misplaced as well :(

I already did that on master yesterday, can be removed from here.

bernt-matthias:
fixed

nsoranzo (approved):

PR #575, opened 2020-09-28
sync github actions workflows with current IUC state

bernt-matthias:
also add badges


mvdbeek:
Thanks @bernt-matthias !

PR #577, opened 2020-09-29
allow larger number of columns for column maker

bernt-matthias:
fixes https://github.com/galaxyproject/tools-devteam/issues/505

idea: 
- use inputs config file
- inject metadata (unfortunatelly that is not in the jsonfile)
- overwrite command line arguments (switched to argparse)
  using json

### DIFF
--- a/tools/column_maker/column_maker.py
+++ b/tools/column_maker/column_maker.py
@@ -5,14 +5,10 @@ the file which is the result of a computation performed on every row in the
 original file. The tool will skip over invalid lines within the file,
 informing the user about the number of lines skipped.
 """
-from __future__ import print_function
 
 import argparse
 import json
 import re
-import sys
-
-assert sys.version_info[:2] >= (2, 4)
 
 parser = argparse.ArgumentParser()
 parser.add_argument('input', type=argparse.FileType('r'), help="input file")
--- a/tools/column_maker/column_maker.py
+++ b/tools/column_maker/column_maker.py
@@ -24,9 +24,9 @@ parser.add_argument('--load_json', default=None, type=argparse.FileType('r'),
                     help="overwrite parsed arguments from json file")
 args = parser.parse_args()
 
+argparse_dict = vars(args)
 if args.load_json:
     json_dict = json.load(args.load_json)
-    argparse_dict = vars(args)
     argparse_dict.update(json_dict)
 
 fh = argparse_dict['input']
--- a/tools/column_maker/column_maker.py
+++ b/tools/column_maker/column_maker.py
@@ -7,29 +7,50 @@ informing the user about the number of lines skipped.
 """
 from __future__ import print_function
 
+import argparse
+import json
 import re
 import sys
 
 assert sys.version_info[:2] >= (2, 4)
 
-inp_file = sys.argv[1]
-out_file = sys.argv[2]
-expr = sys.argv[3]
-round_result = sys.argv[4]
+parser = argparse.ArgumentParser()
+parser.add_argument('input', type=argparse.FileType('r'), help="input file")
+parser.add_argument('output', type=argparse.FileType('wt'), help="output file")
+parser.add_argument('cond', nargs='?', type=str, help="expression")
+parser.add_argument('round', nargs='?', type=str, choices=['yes', 'no'],
+                    help="round result")
+parser.add_argument('columns', nargs='?', type=int, help="number of columns")
+parser.add_argument('column_types', nargs='?', type=str, help="comma separated list of column types")
+parser.add_argument('avoid_scientific_notation', nargs='?', type=str, choices=['yes', 'no'],
+                    help="avoid scientific notation")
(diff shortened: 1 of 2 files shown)


tools/column_maker/column_maker.xml
```diff
@@ -1,164 +1,170 @@
...
-  divmod | exp | float | log | log10 | floor | hex | int | len | long
-
-  max | min | oct | ord | pow | range | reversed
-
-  round | sorted | sqrt | str | sum | type | unichr | unicode |
-
-    </help>
-    <citations />
-</tool>
+<tool id="Add_a_column1" name="Compute" version="1.3.1">
```
bgruening, line 1: the version number should be bumped ...

tools/column_maker/column_maker.xml
```diff
@@ -1,164 +1,170 @@
...
+    <description>an expression on every row</description>
+    <requirements>
+        <requirement type="package" version="2.7.13">python</requirement>
```
bgruening, line 4: can we migrate this to python3

bernt-matthias: seems that the tests work for py 3. 

but there might be functional differences, e.g. integer division now returns float. or is the code using numpy?

tools/column_maker/column_maker.py
```diff
@@ -5,31 +5,48 @@
...
+args = parser.parse_args()
 
-inp_file = sys.argv[1]
-out_file = sys.argv[2]
-expr = sys.argv[3]
-round_result = sys.argv[4]
+if args.load_json:
+    json_dict = json.load(args.load_json)
+    argparse_dict = vars(args)
+    argparse_dict.update(json_dict)
```
hexylena, line 30:
```suggestion
argparse_dict = vars(args)
if args.load_json:
    json_dict = json.load(args.load_json)
    argparse_dict.update(json_dict)
```

bernt-matthias:
@hexylena what is your take on https://github.com/galaxyproject/tools-devteam/pull/577#discussion_r497374267

hexylena:
It could be nice but, isn't the devteam tools policy more or less "maintenance mode" and not making big changes like refactoring python version which might introduce differences, instead only fixing bugs? (Or did I get the wrong impression from somewhere)

bernt-matthias:
I can easily drop the py2to3 commit .. what do you think @bgruening

bernt-matthias:
@hexylena .. I accidentally removed your commit (647bdd7) simplifying the json parsing. Could you restore it?

tools/column_maker/column_maker.py
```diff
@@ -7,29 +7,50 @@
...
+parser.add_argument('column_types', nargs='?', type=str, help="comma separated list of column types")
+parser.add_argument('avoid_scientific_notation', nargs='?', type=str, choices=['yes', 'no'],
+                    help="avoid scientific notation")
+parser.add_argument('--load_json', default=None, type=argparse.FileType('r'),
+                    help="overwrite parsed arguments from json file")
+args = parser.parse_args()
+
+if args.load_json:
+    json_dict = json.load(args.load_json)
+    argparse_dict = vars(args)
```
hexylena, line 33:
```suggestion
argparse_dict = vars(args)
if args.load_json:
    json_dict = json.load(args.load_json)
```

bernt-matthias:
Can we merge this?

tools/column_maker/column_maker.xml
bgruening: I would bump the tool version and port it to python3 - it the results are different that would be fine because its a different version. We do need to make the cut at some point.

tools/column_maker/column_maker.xml
```diff
@@ -1,164 +1,170 @@
...
+            (sed -n '1,1p' '$input' | sed  "s|$|%${header_lines_conditional.header_new_column_name}|" | tr "%" "\t") > header &&
+            sed '1,1d' '$input' > data &&
+        #else:
+            touch header &&
+            ln -s '$input' data &&
+        #end if
+
+        ## inject colums and column_types metadata into inputs json
+        #import json
+        #set inputs_dict = json.load(open($inputs))
```
bgruening, line 19: interesting, why does the path not need any quotes ...

bernt-matthias: True. Will investigate.

bernt-matthias: Its correct the way it is. Essentially its python passing a string variable 

```
inputs="PATH"
open(inputs)
```

but good comment anyway, because I discovered some missing quotes a few lines below.

bgruening (approved):

bgruening:
This looks great to me. Thanks @bernt-matthias! Ready?

bernt-matthias:
> This looks great to me. Thanks @bernt-matthias! Ready?

sure

bgruening:
Merci!

PR #580, opened 2020-10-26
Fix Python2-only syntax in sicer tool

nsoranzo:
Fix:
```
    except Exception, e:
                    ^
SyntaxError: invalid syntax
````

Also cleanup wrapper and script.

### DIFF
--- a/tools/sicer/sicer_wrapper.xml
+++ b/tools/sicer/sicer_wrapper.xml
@@ -1,155 +1,156 @@
 <tool id="peakcalling_sicer" name="SICER" version="0.0.2">
-  <description>Statistical approach for the Identification of ChIP-Enriched Regions</description>
-  <command interpreter="python">sicer_wrapper.py 
-  --bed_file '${input_bed_file}' 
-  #if str( $input_control_file ) != 'None':
-      --control_file '${input_control_file}'
-      --significant_islands_output_file "${significant_islands_output_file}"
-      --islands_summary_output_file "${islands_summary_output_file}"
-      --significant_islands_summary_output_file "${significant_islands_summary_output_file}"
-  #end if
-  ${fix_off_by_one_errors}
-  --dbkey '${input_bed_file.dbkey}'
-  --redundancy_threshold '${redundancy_threshold}'
-  --window_size '${window_size}'
-  --fragment_size '${fragment_size}'
-  --effective_genome_fraction '${effective_genome_fraction}'
-  --gap_size '${gap_size}'
-  --error_cut_off '${error_cut_off}'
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by nsoranzo** [comment]
sicer works fine on 20.05, but biocontainers dependency resolution is broken on 20.09 by https://github.com/galaxyproject/galaxy/commit/f0a55a399cf96cd8c11e3ab8a0d283e9df852088 because the keys in `mulled_resolution_cache` are lowercase while the requirement name is uppercase.

File: tools/sicer/sicer_wrapper.xml [hunk 1]
```diff
@@ -1,155 +1,156 @@
...
-      <output name="island_filtered_output_file" file="test-W200-G600-E0.01-islandfiltered.bed" />
-      <output name="island_filtered_normalized_wig_output_file" file="test-W200-G600-E0.01-islandfiltered-normalized.wig" />
-      <output name="score_island_output_file" file="test_4_test-W200-G600-E0.01.scoreisland" />
-      <output name="output_log_file" file="test_1_output_log_file.contains" compare="contains"/>
-    </test>
-  </tests>
-  <help>
+    <description>Statistical approach for the Identification of ChIP-Enriched Regions</description>
+    <requirements>
+        <requirement type="package" version="1.1">SICER</requirement>
```
mvdbeek, line 4:
```suggestion
        <requirement type="package" version="1.1">sicer</requirement>
```
Nice find, but I think this should be lowercase since the bioconda package and docker images are lowercase (uppercase docker images aren't valid either, xref: https://github.com/galaxyproject/galaxy/pull/9733).

nsoranzo: In general I agree, but on the other hand this has been a nice test case and I don't want to increase the tool version just for that.

mvdbeek: I would say the tool version should be increased, the tempfile handling has changed and it's now Python 3 compatible.

nsoranzo: It's not changing requirements, tool inputs or outputs, so I'd call all these "internal changes", but I could increase the tool version nonetheless if you let me keep the uppercase requirement as a test case :D

nsoranzo: I'm going to merge this, hope that's OK.

drosofff:
Hi @nsoranzo Many thanks for the Python 3 conversion !

I think that the macro.xml file is missing in the repo as well as an import statement in the xml wrapper

drosofff:
Hoops forget it @nsoranzo my bad

PR #581, opened 2020-12-22
Allow non-coordinate-sorted bam input to eXpress

gnaisha:
eXpress requires randomly sorted bams as input, but having the input type set to "bam" causes the bam to be coordinate sorted before running. This PR adds "qname_sorted.bam" and "unsorted.bam" input types to avoid this sorting.

### DIFF
--- a/tools/express/express_wrapper.xml
+++ b/tools/express/express_wrapper.xml
@@ -25,7 +25,7 @@
 	</command>
 	<inputs>
 		<param format="fasta" name="multiFasta" type="data" label="A set of target references (annotation) in multi-FASTA format" help="The multi-FASTA file can also be a fasta file." />
-		<param format="sam,bam" name="bamOrSamFile" type="data" label="Alignments in the BAM or SAM format" help="The set of aligned reads." />
+		<param format="sam,bam,unsorted.bam,qname_sorted.bam" name="bamOrSamFile" type="data" label="Alignments in the BAM or SAM format" help="The set of aligned reads." />
 		<conditional name="additional_params">
             <param name="use_additional" type="select" label="Use Additional Parameters?">
                 <option value="no">No</option>


bernt-matthias:
Thanks for the contribution. The checks are currently not running because github actions changed a bit. They should run again after https://github.com/galaxyproject/tools-devteam/pull/582 is merged. 

Q: Should the sorted bam types be removed? Also a bump of the tool version is necessary.

bernt-matthias:
Can you rebase the PR branch. Then Tests will run.

gnaisha:
Thanks for the fixes to allow the checks to pass.

I tried removing the sorted bam input from the wrapper, but in my local tests it seems that sorted bam is still accepted as input, causing the tool to fail. I've made the change in the wrapper however, since this should be an invalid input type, and will file an issue with Galaxy main regarding disallowing sorted bam in such cases.

I see that there is a tool version in the wrapper, and an eXpress version in tool_dependencies.xml. These are both set to 1.1.1 currently - is this coincidence? I assume I should only bump the tool wrapper version number.

bernt-matthias:
> I see that there is a tool version in the wrapper, and an eXpress version in tool_dependencies.xml. These are both set to 1.1.1 currently - is this coincidence? I assume I should only bump the tool wrapper version number.

You can remove the file `tool_dependencies.xml`

gnaisha:
I've removed that file, and updated eXpress to the most recent version I'm seeing in bioconda.

bernt-matthias:
I should have asked this earlier: What is the exact problem with sorted bam file. The manual suggests that sorting is necessary: 

_If you aligned your reads with Bowtie, your alignments will be properly ordered already. If you used another tool, you should ensure that they are properly sorted_

gnaisha:
Sorry I should have been explicit referring to "sorted" - eXpress does require sorted input, but it needs to be sorted by read name rather than coordinate.

> You can sort your BAM using this command:

    samtools sort -n hits.bam hits.sorted

Galaxy seems to define the "bam" type as being coordinate sorted, and when I used a query-name sorted bam as input to the current eXpress wrapper, Galaxy converted it to a coordinate-sorted bam before running eXpress, causing eXpress to fail. Adding qname_sorted.bam type allows this bam to be used as is, sorted by read name.

bwlang:
Maybe the wrapper should sort on a pipe? For this and other tools that need oddball bags

Brad

On Dec 31, 2020, at 5:39 PM, gnaisha <notifications@github.com> wrote:

﻿
EXTERNAL SENDER


Sorry I should have been explicit referring to "sorted" - eXpress does require sorted input, but it needs to be sorted by read name rather than coordinate.

You can sort your BAM using this command:

samtools sort -n hits.bam hits.sorted


Galaxy seems to define the "bam" type as being coordinate sorted, and when I used a query-name sorted bam as input to the current eXpress wrapper, Galaxy converted it to a coordinate-sorted bam before running eXpress, causing eXpress to fail. Adding qname_sorted.bam type allows this bam to be used as is, sorted by read name.

tools/express/express_wrapper.xml
```diff
@@ -1,7 +1,7 @@
-<tool id="express" name="eXpress" version="1.1.1">
+<tool id="express" name="eXpress" version="1.5.1">
 	<description>Quantify the abundances of a set of target sequences from sampled subsequences</description>
 	<requirements>
-		<requirement type="package" version="1.1.1">eXpress</requirement>
+		<requirement type="package" version="1.5.1">eXpress</requirement>
 	</requirements>
 	<command>
 		express --no-update-check
```
bernt-matthias, line 7:
```suggestion
    #if $bamOrSamFile.ext == "qname_sorted.bam"
        ln -s `$bamOrSamFile` hits.sorted
    #else
        samtools sort -n '$bamOrSamFile' hits.sorted
    #end if
    express --no-update-check
```

Plus: replace `$bamOrSamFile` on the last line of the command block with `hits.sorted` (btw. all input files need to be single quoted).

Maybe in the else branch one could check with `samtools view -H ... | grep -q "@HD\t.*\tSO:unsorted"` if the sam / bam file is sorted before calling sort again.

PR #582, opened 2020-12-26
sync GH actions with IUC

bernt-matthias:
seems that the PR workflow does not work anymore, because of `set-env`. Guess its about time to sync the workflows again.

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -3,6 +3,7 @@ on: [push, pull_request]
 env:
   GALAXY_REPO: https://github.com/galaxyproject/galaxy
   GALAXY_RELEASE: release_20.09
+  MAX_CHUNKS: 4
 jobs:
   # the setup job does two things:
   # 1. cache the pip cache and .planemo
@@ -14,6 +15,11 @@ jobs:
   setup:
     name: Setup cache and determine changed repositories
     runs-on: ubuntu-latest
+    outputs:
+      galaxy_head_sha: ${{ steps.get-galaxy-sha.outputs.galaxy_head_sha }}
+      commit_range: ${{ steps.get-commit-range.outputs.commit_range }}
+      nchunks: ${{ steps.get-chunks.outputs.nchunks }}
+      chunk_list: ${{ steps.get-chunks.outputs.chunk_list }}
     strategy:
       matrix:
         python-version: [3.7]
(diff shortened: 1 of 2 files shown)


.github/workflows/pr.yaml
```diff
@@ -203,7 +238,7 @@ jobs:
     - name: lintr
       run: |
         library(lintr)
-        linters <- with_defaults(line_length_linter = NULL, cyclocomp_linter = NULL)
+        linters <- with_defaults(line_length_linter = NULL, object_usage_linter = NULL)
```
nsoranzo, line 241: Should we keep `cyclocomp_linter = NULL` (added in commit 4e17a8df806c4f37c83874ebf46b97c924586a5a )?

bernt-matthias: Probably yes and we should also add it to the IUC workflow, or?

.github/workflows/pr.yaml
```diff
@@ -396,27 +413,25 @@ jobs:
...
       with:
         path: ~/.cache/pip
-        key: pip_cache_py_${{ matrix.python-version }}_gxy_${{ env.GALAXY_HEAD_SHA }}
+        key: pip_cache_py_${{ matrix.python-version }}_gxy_${{ needs.setup.outputs.galaxy_head_sha }}
     - name: Install Planemo
       run: pip install planemo
     - name: Deploy on testtoolshed
       env:
-        SHED_KEY: ${{ secrets.tts_api_key }}
+        SHED_KEY: ${{ secrets.TTS_API_KEY }}
```
bernt-matthias, line 426: @nsoranzo do you know if the change to uppercase requires an update of the settings?

nsoranzo: Yes, it does need someone with admin privileges on this repo to remove and re-add the TTS and TS API key secrets.

.github/workflows/pr.yaml
nsoranzo: I suppose it would make life easier on tools-iuc too.

bernt-matthias: done. and https://github.com/galaxyproject/tools-iuc/pull/3398

.github/workflows/pr.yaml
bernt-matthias: OK. Given the activity on this repo we can ask for this change after this is merged.

martenson: I doubt this is needed, we have uppercase version in the secrets since June

bernt-matthias: I remember that they are shown upper case even if entered lower case. But we can just try with one of my long open PRs :)

martenson: @bernt-matthias lol, in that case I have no idea what we have. I am ready to change it if it fails though.

bernt-matthias:
Ready to go?

bernt-matthias:
Seems that deployment works https://github.com/galaxyproject/tools-devteam/actions/runs/451752521

PR #590, opened 2021-02-25
Test against virtualenv fix branch

mvdbeek:


nsoranzo:
I think https://github.com/galaxyproject/planemo/issues/1147 was visible only when at least one tool is actually tested.

mvdbeek:
Yup, I meant to test this against my own fork

PR #611, opened 2022-11-02
tabular_to_fasta: Strip title_col input, fix flake8 errors

pvanheus:
While [this PR](https://github.com/galaxyproject/galaxy/pull/14518) addresses the problem where end-of-lines creep into `data_column` inputs in the workflow editor, for older releases of Galaxy (pre-22.05) this still causes a problem for the tabular_to_fasta tool. This PR fixes that while also fixing flake8 errors and using argparse instead of sys.argv for input (so that if errors with file I/O occur they are more readable).

### DIFF
--- a/tools/tabular_to_fasta/tabular_to_fasta.py
+++ b/tools/tabular_to_fasta/tabular_to_fasta.py
@@ -1,10 +1,10 @@
 #!/usr/bin/env python
 """
-Input: fasta, minimal length, maximal length
+Input: fasta, title columns (comma separated), sequence column
 Output: fasta
-Return sequences whose lengths are within the range.
+Return tabular format sequences converted to FASTA
 """
-import os
+import argparse
 import sys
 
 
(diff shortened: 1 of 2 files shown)


tools/tabular_to_fasta/tabular_to_fasta.py
```diff
@@ -13,53 +13,56 @@ def stop_err(msg):
...
+    if title_col is None or title_col == 'None' or seq_col is None or seq_col == 'None':
         stop_err("Columns not specified.")
     try:
         seq_col = int(seq_col) - 1
-    except:
+    except ValueError:
         stop_err("Invalid Sequence Column: %s." % str(seq_col))
 
-    title_col_list = title_col.split(',')
+    title_col_list = title_col.strip().split(',')
```
nsoranzo, line 32: Given the chat with Marius on Matrix, revert this?

```suggestion
    title_col_list = title_col.split(',')
```

PR #612, opened 2023-02-22
Fix kraken2tax requirements and add test

mvdbeek:


nsoranzo:
Should this tool be migrated to tools-iuc?

mvdbeek:
What's the advantage ? I think these low-priority, not-frequrently-updated tools can live here

mvdbeek:
I see now what the advantage is, but I think it's a nice motivation for https://github.com/galaxyproject/planemo-ci-action/pull/43 and community tool repos in general.

nsoranzo:
Also we don't test this repo before a new Galaxy release, which won't be "fixed" by reusable workflows.

mvdbeek:
> Also we don't test this repo before a new Galaxy release, which won't be "fixed" by reusable workflows.

We do, it's in the release process. I've just not done it recently.

mvdbeek:
Actually, neither IUC nor devteam repo tests are part of the "official" release process. We don't do that for e.g bjoern's tools either, i don't think that should be a criterium one way or the other.

nsoranzo (approved):

PR #623, opened 2024-03-12
Add bio.tools for t2t report

EngyNasr:
Hi all,

This PR links the tools to its bio.tools entry: https://bio.tools/t2t_report
It will be useful to get EDAM ontology annotations

Engy


bernt-matthias (commented):
Tests won't run unless https://github.com/galaxyproject/tools-devteam/blob/4196e9b12ab23035e6cddef417a0282d2d49de11/.tt_skip#L10 is removed. 

In general, can you obtain for the tools in devteam that we want to touch here statistics on how much they are used on usegalaxy.eu?

PR #624, opened 2024-03-15
Update cd_hit_dup.xml

Minamehr:
Hi- The bio.tools ID (Link: https://bio.tools/cd_hit_dup) has been added to the tool wrapper.


bernt-matthias:
You have chosen one of the difficult cases :) 

- In order to get the tests running we need to remove this line: https://github.com/galaxyproject/tools-devteam/blob/4196e9b12ab23035e6cddef417a0282d2d49de11/.tt_skip#L23
- `cd-hit-dup` will be included here .. at some point .. https://github.com/galaxyproject/tools-iuc/pull/3328

PR #628, opened 2024-10-03
fasta_compute_length: Add required file for Pulsar deployments and bump python to 3.12

fubar2:
Bump to python3.12 from 3.7
Add the python script as a required file for Pulsar.


tools/fasta_compute_length/fasta_compute_length.xml
```diff
@@ -1,8 +1,11 @@
-<tool id="fasta_compute_length" name="Compute sequence length" version="1.0.3" profile="16.04">
+<tool id="fasta_compute_length" name="Compute sequence length" version="1.0.3+galaxy1" profile="22.04">
```
nsoranzo, line 1: Maybe a patch version number bump makes more sense, as there isn't an underlying tool version to align to?

```suggestion
<tool id="fasta_compute_length" name="Compute sequence length" version="1.0.4" profile="22.04">
```

nsoranzo (dismissed):

tools/fasta_compute_length/fasta_compute_length.xml
```diff
@@ -1,8 +1,11 @@
...
     <description></description>
     <requirements>
-        <requirement type="package" version="3.7">python</requirement>
+        <requirement type="package" version="3.12">python</requirement>
     </requirements>
+    <required_files>
+        <include path="fasta_compute_length.py"/>
```
wm75, line 7: I'm confused: including this file is certainly nice and recommended best pratice nowadays as I just learnt :-),
but isn't the *important* bit to include utils/fasta_to_len.py because that's the one that pulsar will not be able to figure out from the xml?

fubar2: Good catch @wm75! Got no way to test this in pulsar so all advice appreciated.

So probably need both if pulsar has to copy everything needed into the remote execution environment.

Oh - just realised - even easier - how about just consolidating that dependency - which is really the script - into the script to get rid of the utils complication? 

So now, should work for pulsar as is - no need for utils

nsoranzo (approved):

wm75 (approved):

PR #629, opened 2024-11-22
add galaxy codebase license

fubar2:

PR #630, opened 2024-11-24
Create LICENSE

bgruening:
Urgs, that is probably the worst repo without a license.

@nsoranzo any idea? Pinging all people?


nsoranzo:
I think #629 is better because most of the code here was moved out of the main galaxy repo, with its Academic license.

PR #631, opened 2025-01-23
Add "use_header_names" option to histogram

hechth:
The histogram plotting tool doesn't use the header to display the column selection, which is quite inconvenient.

In case there is no header, I think this should still work? Or should we add a boolean to check whether the file has a header?


hechth:
@jdavcs or @nsoranzo could someone review and merge this please? Thanks a lot!

jdavcs (approved):
I think this is fine. Works for me. @nsoranzo can you double-check, please? (I don't review tool edits often, so I may be missing something)

PR #633, opened 2025-02-03
Update scatterplot.xml to use header names

### LINKED ISSUE #632: Scatterplot: select the columns based on colnames rather than position in the table
Hi! Would it be possible to change in [scatterplot](https://github.com/galaxyproject/tools-devteam/tree/main/tools/scatterplot) the numerical columns choice based on the colnames rather than the position in the table? Especially in the larger tables it gets impractical to find the correct columns position :)

Image

Thanks for the consideration!

hechth:
closes #632

### DIFF
--- a/tools/scatterplot/scatterplot.xml
+++ b/tools/scatterplot/scatterplot.xml
@@ -17,8 +17,8 @@
   </command>
   <inputs>
     <param name="input" type="data" format="tabular" label="Dataset" help="Dataset missing? See TIP below"/>
-    <param name="col1" type="data_column" data_ref="input" numerical="True" label="Numerical column for x axis" />
-    <param name="col2" type="data_column" data_ref="input" numerical="True" label="Numerical column for y axis" />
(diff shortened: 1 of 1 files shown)


hechth:
ping @mvdbeek @martenson @bernt-matthias @bgruening please merge

tools/scatterplot/scatterplot.xml
```diff
@@ -1,9 +1,15 @@
-<tool id="scatterplot_rpy" name="Scatterplot" version="1.0.3">
+<tool id="scatterplot_rpy" name="Scatterplot" version="1.0.3+galaxy0">
   <description>of two numeric columns</description>
   <requirements>
     <requirement type="package" version="1.9">numpy</requirement>
     <requirement type="package" version="2.8.5">rpy2</requirement>
+    <requirement type="package" version="3.0">libgfortran</requirement>
```
martenson, line 6: why does this need fortran compiler?

hechth: Some of the R packages seem to need this - this was probably previously contained in the container or base image or came with R or something but it basically complains that it can't find libgfortran.so if this isn't added :/

